PR #43, opened 2015-02-24
Add chunkBy and divideInto

latkin:


dsyme:
Please use chunkBySize as interim since that's my latest preferred name, thanks. (It's either that or plain "chunk" I think)

latkin:
Done

dsyme:
Thanks!

PR #57, opened 2016-03-01
Mention interaction with implicit flexibility on use of function-typed values

dsyme:


smoothdeveloper:
@dsyme & others, I think we should use the ```fsharp backtick thing around codeblocks to make them look nicer!

dsyme:
@smoothdeveloper Please send a PR

PR #71, opened 2016-03-06
add section about the guaranteed order of results

mexx:
as suggested by Don in https://github.com/fsharp/FSharpLangDesign/issues/47#issuecomment-190665773


smoothdeveloper:
@mexx could you also mention `cartesian` under proposed names? Thanks!

mexx:
@smoothdeveloper not in this PR as it's already merged ;) btw you can easily do it yourself here on the page. That's how I've proposed this PR.

PR #74, opened 2016-03-09
Collect all the suggested names

mexx:


RFCs/FS-1004-result-type.md
```diff
@@ -25,7 +25,14 @@ The type should be added in `FSharp.Core` library
...
 
-The name of the operator is `Result`
+The name of the type and union cases has not been finalized. Some of the suggestions are
+
+| Type name | Success case  | Failure case |
+| --------- | ------------- | ------------ |
+| Result    | Success       | Failure      |
+| Result    | Success       | Error        |
+| Result    | OK            | Error        |
+| Result    | OK            | Bad          |
```
vasily-kirichenko: You forget `Ok` and `Err`.

mexx: It was actually the reason why I've started the table ;)
However it slipped through. Now in.

PR #80, opened 2016-04-05
Record resolution of issues

dsyme:


RFCs/FS-1004-result-type.md
```diff
@@ -17,34 +17,19 @@ it's useful to have a fsharp type who contains the result on success or the erro
...
-The standard range of testing considerations for F# library types apply.
+### Resolved Questions
 
-### Open Questions
-
-- namespace
-- declare as struct
-- compiled name
-- require additional functions?
+* No additional functions or methods will be supported in FSharp.Core.   These may be added in a community library or user code.
```
dtchepak: Any reason why there will be no additional functions in FSharp.Core?  `map`, `mapError`, `bind` and `fold` would be pretty useful.

neoeinstein: This decision follows along with the general guidelines for the F# Core library and other types that have been provided.

neoeinstein:
Looks good for merge.

RFCs/FS-1004-result-type.md
dtchepak: Thanks @neoeinstein. Are these guidelines available somewhere? (I'm new to Core dev stuff so sorry if I've missed an obvious link :) ).

ploeh: Like other people, both here and [in the feature discussion thread](https://github.com/fsharp/FSharpLangDesign/issues/49#issuecomment-206318285), I think it'd be an error not to include standard functions like at least `map` and `bind`, and preferably also a computation expression builder.

# Community code

Relying on community code for general-purpose functionality like this, is, in my opinion, not the optimal solution.

Essentially, I can identify two directions that community-supplied packages could take:

- Fine-grained packages
- Coarse-grained packages

Both options have disadvantages, although among these alternatives, I prefer fine-grained packages.

## Fine-grained packages

Although fine-grained packages are my preferred choice among these alternatives, they come with disadvantages. A [recent example from the JavaScript/Node community](http://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos) demonstrates how the use of packages that should have been core library functions puts an entire community at risk.

In addition to that, depending on myriads of fine-grained packages creates maintenance friction for application developers.

## Coarse-grained packages

Coarse-grained packages, on the other hand, come with the disadvantage that they force you to pull in 'everything and the kitchen-sink'. In order to get the benefit of e.g. `Result.map` or `Result.bind`, you'll need to pull in a big library that exports tons of other, undesired functionality.

Perhaps you'll even end up with multiple big libraries, but they'll provide overlapping features. Let me describe a hypothetical example:

You pull in _FooLib_ because you need `Result.bind`. You also, previously, pulled in _BarLib_, because you need `Async.map`.

I'd be surprised if any coarse-grained library of general-purpose F# functions don't define standard functions like `flip`, `curry`, `uncurry`, etcetera, but these are probably fairly benign because they are unlikely to be different.

It's not a stretch, however, to imagine that both libraries export a module to handle pairs (two-element tuples). Both libraries might define a `Tuple2.map` function, but what does this function do?

_FooLib_ could define it as

```F#
let map f (x, y) = f x, y
```

while _BarLib_ might define it as

```F#
let map f (x, y) = x, f y
```

Imagine the hours of frustration and troubleshooting something like that might create in a user code base.

(Incidentally, Haskell defines `fmap` for pairs as mapping the second element, so equivalent to the second option... That's probably the most intuitive implementation for pairs, but as always, 'intuitive' is subjective.)

Another source of confusion could arise if both coarse-grained libraries define a computation expression for `Result` types, but with slightly different feature sets. Again, as an application user, you may be writing code inside a `result` computation expression, and then wonder why it doesn't compile. After all, you have similar code in a different part of your code base, and it compiles, so why does the compiler complain about your use of `return!` in a new piece of code that you're writing? After hours of troubleshooting, you finally discover that you're using the 'wrong' implementation of `result`.

# User code

The other alternative hinted at here is that `map` and `bind` functions can be provided in user code. Granted, they are trivial, but now, instead of one single implementation, we'll have thousands of implementations. How is that helpful to the F# community?

# FSharp.Core

For the above reasons, I think that functions like `Result.map` and `Result.bind` should be defined in FSharp.Core, just like `Option.bind`, `List.map`, etcetera.

On the other hand, I think FSharp.Core should be exclusively shipped as a NuGet package, and never as a GAC'ed assembly. This would enable the community to release new features and updates much faster than when it follows the Visual Studio release cadence.

It would also resolve the messed-up situation with dependency management related to this particular library. As an example, adding some NuGet packages to a Visual Studio solution can actually _downgrade_ your FSharp.Core dependency.

dsyme: @ploeh Thanks for the comments and the reasoning. See also my comment here: https://github.com/fsharp/FSharpLangDesign/issues/49#issuecomment-206822451

FWIW I think "error" is a bit strong :) Perhaps "sub-optimal" would be better?  In particular defining the type in FSharp.Core has utility for interop between F# components, which is one primary concern being addressed here.

ploeh: @dsyme Thank you for responding. I saw that other comment, and find it entirely reasonable :smile: 

I agree that _sub-optimal_ would be more precise than _error_. Please accept my apologies if my choice of word(s) caused grief.

PR #81, opened 2016-04-05
Create RFC for FS-1011: Warn when recursive function is not tail-recursive

neoeinstein:


forki:
Did you remove the template?

neoeinstein:
Not the core template, no. This was one of pre-created placeholders for this RFC number (there are also ones for 1012 and 1013)

forki:
I see. Ignore me.

RFCs/FS-1011-Warn-on-recursive-without-tail-call.md
```diff
@@ -0,0 +1,85 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+* Use the reserved `tailcall` keyword rather than an attribute
+* Apply the inspection to all recusive functions
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* What warning level should the warning be issued at?
```
dsyme: There are a bunch of unresolved issues here, I'll add them to the list

dsyme:
Thanks.  I remember finding a host of other issues when trying to implement this - I'll try to dig out the work and augment the RFC appropriately (one thing I like about the RFC process is it gives a proper place to record issues!)

Issue #49, opened 2016-02-14
[RFC FS-1004 Discussion] Result type


enricosada:
This issue is used to track discussions of [F# RFC FS-1004 - "Result type"](https://github.com/fsharp/fslang-design/blob/master/FSharp-4.1/FS-1004-result-type.md). Please discuss in thread below (if necessary)

forki:
https://github.com/Microsoft/visualfsharp/pull/964 is proposing to add

    /// <summary>Helper type for error handling without exceptions.</summary>
    [<StructuralEquality; StructuralComparison>]
    [<CompiledName("FSharpResult`2")>]
    type Result<'T1,'T2> = 
    | Success of 'T1 
    | Error of 'T2
 
Is that really the type we want for Results? Chessie is using:

    /// Represents the result of a computation.
    type Result<'TSuccess, 'TMessage> = 
    /// Represents the result of a successful computation.
    | Ok of 'TSuccess * 'TMessage list
    /// Represents the result of a failed computation.
    | Bad of 'TMessage list

If we add a specialized type to FSharp.Core then we should really think about it's usage. the proposed type is maybe too generic and basically only an alias for Choice.

enricosada:
I think it's good to have a shared (in whole f# ecosystem) Success/Failure type.
But I think is wrong to add more functions in FSharp.Core.
External libraries like Chessie can add additional functions using the Result type.
Maybe the only function i'd like to add is `Result.ofChoice: Choice1Of2 -> Result` but maybe it's not needed

@swlaschin do you have a feedback to add?

wallymathieu:
@forki Compared to Chessie you could either create a specialised type or an alias : ```RopResult<'TSuccess, 'TMessage> = Result<'TSuccess*'TMessage list,'TMessage list>```
The difference with Choice is that this type tells you that one of the states is a failure and the other is a success. Choice1Of2 does not read like Success.

forki:
Yes, but IMHO this alone is not useful enough to justify this type in the FSharp.Core

wallymathieu:
@swlaschin has provided an example of railway oriented programming, if you use an alias for the type in [Railway-Oriented-Programming-Example](https://github.com/wallymathieu/Railway-Oriented-Programming-Example/tree/how_would_it_look_with_alias) you need to add a few more type signatures.

haf:
The success case should not contain a list of messages.

forki:
[...]

PR #83, opened 2016-04-07
Some options for Result module functions (#49)

dtchepak:
Starting point for discussing functions for possible inclusion with `Result` type.

### DIFF
--- a/RFCs/FS-1004-result-type.md
+++ b/RFCs/FS-1004-result-type.md
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
 - declare as struct
 - compiled name
 - require additional functions?
+
+### Proposal for additional functions
+
+Minimal:
+
+```fsharp
+bind : ('TSuccess -> Result<'T, 'TError>) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+map : ('TSuccess -> 'T) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+```
+
+Nice-to-have:
+
+```fsharp
+fold : ('TSuccess -> 'T) -> ('TError -> 'T) -> Result<'TSuccess, 'TError> -> 'T
+mapError : ('TError -> 'T) -> Result<'TSuccess, 'TError> -> Result<'TSuccess, 'T>
+bimap : ('TSuccess -> 'T) -> ('TError -> 'U) -> Result<'TSuccess, 'TError> -> Result<'T, 'U>
+// `bimap` or `mapBoth`?
+```
+
+Low priority -- should we consider the following to get some parity with the `Option` module?
+
+```fsharp
+iter : ('TSuccess -> unit) -> Result<'TSuccess, 'TError> -> unit
+toArray : Result<'TSuccess, 'TError> -> 'TSuccess []
+toList : Result<'TSuccess, 'TError> -> 'TSuccess list
+get : Result<'TSuccess, 'TError> -> 'TSuccess option
+getError : Result<'TSuccess, 'TError> -> TError option
+```
+
+Builder (possibly as a separate RFC to add an `Option` and `Result` builder?):
+
+```fsharp
+type ResultBuilder() =
+    member x.Return a = Success a
+    member x.Bind(result, f) = Result.bind f result
+    member x.ReturnFrom result = result
+```
+
+
--- a/RFCs/FS-1004-result-type.md
+++ b/RFCs/FS-1004-result-type.md
@@ -51,39 +51,21 @@ The standard range of testing considerations for F# library types apply.
 
 ### Proposal for additional functions
 
-Minimal:
+Uncontentious:
 
 ```fsharp
-bind : ('TSuccess -> Result<'T, 'TError>) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
-map : ('TSuccess -> 'T) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+bind : ('T -> Result<'U, 'TError>) -> Result<'T, 'TError> -> Result<'U, 'TError>
+map : ('T -> 'U) -> Result<'T, 'TError> -> Result<'U, 'TError>
+mapError : ('TError -> 'U) -> Result<'T, 'TError> -> Result<'T, 'U>
 ```
 
-Nice-to-have:
+A result builder should be a separate RFC which covers adding builders for both `Option` and `Result`.
+
+Suggested:
 
 ```fsharp
-fold : ('TSuccess -> 'T) -> ('TError -> 'T) -> Result<'TSuccess, 'TError> -> 'T
-mapError : ('TError -> 'T) -> Result<'TSuccess, 'TError> -> Result<'TSuccess, 'T>
-bimap : ('TSuccess -> 'T) -> ('TError -> 'U) -> Result<'TSuccess, 'TError> -> Result<'T, 'U>
-// `bimap` or `mapBoth`?
+attempt : (unit -> 'a) -> Result<'a,exn>
 ```
 
-Low priority -- should we consider the following to get some parity with the `Option` module?
-
-```fsharp
-iter : ('TSuccess -> unit) -> Result<'TSuccess, 'TError> -> unit
-toArray : Result<'TSuccess, 'TError> -> 'TSuccess []
-toList : Result<'TSuccess, 'TError> -> 'TSuccess list
-get : Result<'TSuccess, 'TError> -> 'TSuccess option
-getError : Result<'TSuccess, 'TError> -> TError option
-```
-
-Builder (possibly as a separate RFC to add an `Option` and `Result` builder?):
-
-```fsharp
-type ResultBuilder() =
-    member x.Return a = Success a
-    member x.Bind(result, f) = Result.bind f result
-    member x.ReturnFrom result = result
-```
 
 


RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
...
+bimap : ('TSuccess -> 'T) -> ('TError -> 'U) -> Result<'TSuccess, 'TError> -> Result<'T, 'U>
+// `bimap` or `mapBoth`?
+```
+
+Low priority -- should we consider the following to get some parity with the `Option` module?
+
+```fsharp
+iter : ('TSuccess -> unit) -> Result<'TSuccess, 'TError> -> unit
+toArray : Result<'TSuccess, 'TError> -> 'TSuccess []
+toList : Result<'TSuccess, 'TError> -> 'TSuccess list
```
ploeh: For completeness, how about adding `toSeq` as well?

```F#
toSeq : Result<'TSuccess, 'TError> -> 'TSuccess seq
```

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
...
+get : Result<'TSuccess, 'TError> -> 'TSuccess option
```
ploeh: May I suggest `getSuccess`, in order to mirror `getError`?

The suggested `get` function isn't equivalent to `Option.get` anyway, because `Option.get` is unsafe. I do, indeed, prefer the type suggested here, but think that `getSuccess` is a more explicit name.

mexx: :+1: for mirroring the names.

Another suggestions for the names instead of using `get` we could use `try` like almost all functions returning `option` types do. Just my 2c.

ploeh: `trySuccess` and `tryError`. I like those :+1:

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
 - declare as struct
 - compiled name
 - require additional functions?
+
+### Proposal for additional functions
+
+Minimal:
+
+```fsharp
+bind : ('TSuccess -> Result<'T, 'TError>) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
```
vasily-kirichenko: Why such lengthy type variable names? What about 
```fsharp
bind : ('a -> Result<'b, 'e>) -> Result<'a, 'e> -> Result<'b, 'e>`
```
?

RFCs/FS-1004-result-type.md
vasily-kirichenko: What about `toOption` as it's called everywhere?

ploeh: What would the type of `toOption` be?

vasily-kirichenko:
2 cents from a guy (me) who have used `Choice`-based error handling for about 2 years.

* Another must have function:
```fsharp
attempt : (unit -> 'a) -> Result<'a, exn>
```
* The computation expression builder must be complete, with all the members implemented, we can simply copy paste it from here https://github.com/jack-pappas/ExtCore/blob/master/ExtCore/Control.fs#L852
* Also `asyncResult` CE must be added, we can copy paste it from here as well https://github.com/jack-pappas/ExtCore/blob/master/ExtCore/Control.fs#L1478

The above implementations are good and stable, no risk.

_Or_ we can put the `Result` type and tons of CEs and functions in a separate NuGet package. It's clear that such a package will be created anyway because people do want much more functionality around `Result` than MS wants to put into FSharp.Core. 

I don't see any advantages in placing `Result` into FSharp.Core because it's unlikely it will be used inside that library for anything ('coz it uses exceptions for error handling).

The cool thing about putting this stuff into a NuGet package is that it can evolve very quickly, without such a bureaucraty process as we are in right now.

RFCs/FS-1004-result-type.md
vasily-kirichenko: `toOption : Result<'a, 'e> -> 'a option`

https://github.com/jack-pappas/ExtCore/blob/master/ExtCore/Pervasive.fs#L1131

Also take a look at `ofOption`, `ofOptionWith` and a bunch of other functions there. I don't think we should invent the wheel here.

mexx:
The only advantage I can see over putting the `Result` type into a separate library is the ability to use the `Result` type for error handling inside FSharp.Core in the future. Putting it into a separate library would drive this option almost towards being impossible.

vasily-kirichenko:
@mexx yes, but I think nobody is going to add exception based API => Result base API wrappers into FSharp.Core anyway. For example, I'd like to see such wrappers for just everything in .NET, starting from `System.IO` and `System.Net`, like:
```fsharp
module IO =
  type Error = Error of exn

  module File =
    let exists (path: FilePath) : Result<bool, Error> = 
      Result.attempt (fun _ -> System.IO.File.Exists path)
...
result {
  let! exists = IO.File.exists file
  ...
}
```

toburger:
@vasily-kirichenko `attempt` is exception based. `Result<'a, 'b>` and the module functions introduced are generic.

RFCs/FS-1004-result-type.md
ploeh: Then what would you call the function with the type `Result<'TSuccess, 'TError> -> TError option`?

vasily-kirichenko:
@toburger I don't get your point, sorry. Could you elaborate?

RFCs/FS-1004-result-type.md
vasily-kirichenko: I have no idea. Never need such a function. Could you give a usage example?

toburger:
@vasily-kirichenko

you said:

> I don't see any advantages in placing Result into FSharp.Core because it's unlikely it will be used inside that library for anything ('coz it uses exceptions for error handling).

The Result type and the proposed functions introduced don't contain any exception type. They clearly state, that there is a Success of something or a Error of something (not an Error of exn).

vasily-kirichenko:
@toburger still don't get you point. OK, could you sketch a code which wraps `File.Exists` into `Result`?

toburger:
You are to picky on the File.Exists example. Yes, for this example it is best to return the exception wrapped in the Error type (because it "returns" an exception in case of an error).

But it is not always the case that you want to return an exception. If you look at Scott Wlashin's article [Railway Oriented Programming](https://fsharpforfunandprofit.com/rop/) it makes often sense to wrap "something" in an error, which you can handle afterwards.

vasily-kirichenko:
@toburger I did wrap the exception into `type Error = Error of exn`. You could wrap it into a more abstract DU case like 
```fsharp
type Error = Error of FileName * error: string`
```
or even
```fsharp
type Error = FileExistsError of FileName | ... other error cases
```
But the problem with such an approach is that the actual exception and stack trace is effectively erased and when you investigate a log file it's impossible to say where exactly error like `ErrorExistsError ("foo", "file not found")` was raised. When I started to use choice-base error handling, I did erase exceptions, but brought them back very quickly.

toburger:
@vasily-kirichenko your aproach makes total sense for exception based error handling. I'm not arguing against that.
But my point is, that a Result type is not necessarily exception based and it does make sense to include such a type in FSharp.Core.

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
...
+```fsharp
+iter : ('TSuccess -> unit) -> Result<'TSuccess, 'TError> -> unit
+toArray : Result<'TSuccess, 'TError> -> 'TSuccess []
+toList : Result<'TSuccess, 'TError> -> 'TSuccess list
+get : Result<'TSuccess, 'TError> -> 'TSuccess option
+getError : Result<'TSuccess, 'TError> -> TError option
+```
+
+Builder (possibly as a separate RFC to add an `Option` and `Result` builder?):
+
```
dsyme: Yes, I think this should best be an separate RFC - if we add one we should add both.

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
...
+bind : ('TSuccess -> Result<'T, 'TError>) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+map : ('TSuccess -> 'T) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+```
+
+Nice-to-have:
+
+```fsharp
+fold : ('TSuccess -> 'T) -> ('TError -> 'T) -> Result<'TSuccess, 'TError> -> 'T
+mapError : ('TError -> 'T) -> Result<'TSuccess, 'TError> -> Result<'TSuccess, 'T>
+bimap : ('TSuccess -> 'T) -> ('TError -> 'U) -> Result<'TSuccess, 'TError> -> Result<'T, 'U>
```
dsyme: Just to mention there is no precedent in FSharp.Core for something with this name.  I suspect it's use is very rare for the Result type, and code would be clearer using ``map`` then ``mapError``.

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
...
+
+```fsharp
+bind : ('TSuccess -> Result<'T, 'TError>) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+map : ('TSuccess -> 'T) -> Result<'TSuccess, 'TError> -> Result<'T, 'TError>
+```
+
+Nice-to-have:
+
+```fsharp
+fold : ('TSuccess -> 'T) -> ('TError -> 'T) -> Result<'TSuccess, 'TError> -> 'T
```
dsyme: I think this would be used very, very rarely, and code would be much clearer with an explicit match.

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,42 @@ The standard range of testing considerations for F# library types apply.
...
+mapError : ('TError -> 'T) -> Result<'TSuccess, 'TError> -> Result<'TSuccess, 'T>
+bimap : ('TSuccess -> 'T) -> ('TError -> 'U) -> Result<'TSuccess, 'TError> -> Result<'T, 'U>
+// `bimap` or `mapBoth`?
+```
+
+Low priority -- should we consider the following to get some parity with the `Option` module?
+
+```fsharp
+iter : ('TSuccess -> unit) -> Result<'TSuccess, 'TError> -> unit
```
dsyme: I don't think ``iter`` adds much over an explicit match. It is also "dangerous" as it drops information if a failure occurred.

RFCs/FS-1004-result-type.md
dsyme: I don't really like any of these ``toList``, ``toArray`` or ``toSeq``, partly because information is lost if a ``Error`` case occurred (all you get is an empty collection).  An explicit match would be clearer in these cases and force the user to do something explicit with the ``Error`` case.

RFCs/FS-1004-result-type.md
dsyme: I suppose the use case might be if the ``Ok`` case were carrying type ``unit``.  

``someAction |> Result.toErrorOption``

That's ok of it only accepts things of type ``Result<unit,T>``.  But I don't really like it if it accepts generic success value: in that case it feels like it needs to have ``ignore`` in the name.

dsyme:
@vasily-kirichenko 

> I don't see any advantages in placing  Result  into FSharp.Core 

The main point of the RFC is to ensure all F# components interested in using a ``Result`` type in their API are using the same compatible type definition, without needing to add a new dependency.

Cheers
Don

RFCs/FS-1004-result-type.md
dsyme: @vasily-kirichenko The F# guidelines clearly recommend ``'t`` named type variables, and it's standard throughout FSharp.Core.  

But I would prefer ``Result<'T,'TError`>, I don't think ``'TSuccess`` adds anything over ``'T``

RFCs/FS-1004-result-type.md
dtchepak: @dsyme, yes not a big difference between an explicit match, but I do find fold functions useful (and use them very frequently in my code):
```fsharp
textBox.value
|> validate
|> function 
    | Success x -> Some x
    | Error _ -> None

// or
textBox.value
|> validate
|> Result.fold Some (constant None)
```

It can also be useful to partially apply a fold function at times.

We could also provide this instead of a bunch of other proposed functions like `ofOption` or `toSeq`, so these can be defined inline as needed (which avoids some of the problems mentioned about inadvertently dropping values):

```fsharp
let ofOption = Result.fold Some (constant None)
let toSeq = Result.fold (Seq.singleton) (constant Seq.empty)
```

RFCs/FS-1004-result-type.md
dtchepak: I prefer short type variable names (`Result<'a,'e>`), but was skimming MSDN and trying to follow conventions there. (I obviously inferred them incorrectly :) ).

dtchepak:
Added a new commit aggregate review comments.

Everyone seems happy with `bind`, `map` and `mapError`. Added `attempt`. Dropped everything else in favour of explicit matches (although I'd still like [`fold`](https://github.com/fsharp/FSharpLangDesign/pull/83#discussion_r59104056) as a replacement for these 😄 ).

RFCs/FS-1004-result-type.md
```diff
@@ -48,3 +48,24 @@ The standard range of testing considerations for F# library types apply.
...
+map : ('T -> 'U) -> Result<'T, 'TError> -> Result<'U, 'TError>
+mapError : ('TError -> 'U) -> Result<'T, 'TError> -> Result<'T, 'U>
+```
+
+A result builder should be a separate RFC which covers adding builders for both `Option` and `Result`.
+
+Suggested:
+
+```fsharp
+attempt : (unit -> 'a) -> Result<'a,exn>
```
dsyme: There's no precedent for naming this kind of function in FSharp.Core. ``attempt`` seems OK but we should stop and consider, since it will set a precedent for all future F# code.

We should probably add ``Option.whatever`` as well, with type ``(unit -> 'T)  -> 'T option``.

dsyme: We should consider if ``exn`` should be ``ExceptionDispatchInfo`` here. That preserves stack traces in a compositional way.

There is also the converse, ``Result<'T,ExceptionDispatchInfo> -> 'T`` which re-raises the exception.

dtchepak: Wouldn't `Option.whatever` cause the loss of the exception information, similar to concerns over `Result<'a,e'> -> 'a option`? Might be better to explicitly drop it with `Result.whatever |> function | Success x -> Some x | _ -> None`.

neoeinstein: I think that `ExceptionDispatchInfo` would be a good improvement in this case. It's not something that I had heard of until now.

neoeinstein:
Since there are likely many existing implementations which use `Choice<'a,'b>` as their `Result`-like construct, would it be valuable to provide `toChoice : (result : Result<'a,'b>) -> Choice<'a,'b>` and `ofChoice : (choice : Choice<'a,'b>) -> Result<'a,'b>` functions?

RFCs/FS-1004-result-type.md
vasily-kirichenko: > ExceptionDispatchInfo cannot be serialized and is not intended to cross application domain boundaries.

It's not good at all.

dsyme:
@neoeinstein I think interoperating with ``Choice`` will be relatively rare in the long term, so I'd prefer not to include choice conversions in the to/from functions in FSharp.Core.  For example, we don't have Option.toChoice or Option.orChoice.

RFCs/FS-1004-result-type.md
dsyme: @vasily-kirichenko There's a long discussion about the future of exception serialization on .NET Core [here](https://github.com/dotnet/coreclr/issues/2715).  I think we need to treat it as an orthogonal issue.

dsyme:
My apologies for the delay in merging this

PR #104, opened 2016-06-29
Minor correction to cross definition

ckimes89:
I believe `yM` should be `y1` - looks like a simple typo or maybe copy+paste error.


mexx:
@ckimes89 thanks for catching it, what ever it was on my side, definitely no intent ;)

PR #131, opened 2016-09-24
Add FS-1025.

kurtschelfthout:
Improve record type inference.


dsyme:
@kurtschelfthout Please also update https://github.com/fsharp/FSharpLangDesign/blob/master/README.md (currently an index maintained by hand).  Or else we auto-generate that.  Or delete that (though a visual presentation of accepted/closed/active/... RFCs is very useful!)

kurtschelfthout:
yes, will do.

PR #135, opened 2016-10-25
Update struct RFCs

neoeinstein:
Add some more data about struct record and DUs when consumed by prior versions of the F# compiler. Notably that v4.0 and earlier will not properly recognize struct DUs and records as value types, thus producing unverifiable code.


dsyme:
Excellent addition, thank you

PR #136, opened 2016-10-29
First cut of Allow params dict as method argument

kurtschelfthout:
Not finished, but I think useful...


dsyme:
Superb, thank you!!

PR #137, opened 2016-10-30
Move F# 4.1 RFCs to dedicated folder

cartermp:
Since F# 4.1 is complete in terms of language features, let's move the relevant RFCs to their own folder.  I've also updated the struct single case union RFC to include multicase unions.


dsyme:
@cartermp I think there are lots of incoming links to these in many places - do we care?

cartermp:
That's a good point, yeah.  I can update the F# 4.1 blog post, which would address a few of those.  Luckily we don't link to here from the documentation (yet), so it's most likely just lingering links.  I think it's good to have this separation just to make it clear what's in F# vNext and what's in previous version of F#.

smoothdeveloper:
@cartermp / @dsyme it seems the proper way to link then would be to point to a revision in time (although it won't have most recent edits, but those aren't supposed to change much after the actual shipping of the code in a release).

I'm refering to the github uris with the commit id in them.

I wish github would make it easier to get that uri because it takes me 3 clicks (including scanning to pick the right file) to get such link from browsing the file.

adamchester:
@smoothdeveloper https://help.github.com/articles/getting-permanent-links-to-files/

dsyme:
@cartermp Merged.  Could you fix up the README please? thanks

cartermp:
Yup, will do!

PR #138, opened 2016-10-31
FS-1026: fix a typo

ForNeVeR:


neoeinstein:
Good catch. Thanks for the correction.

PR #148, opened 2016-11-22
Updates with approved/denied functions

neoeinstein:


dsyme:
Thank you, excellent work

PR #149, opened 2016-11-23
Update FS-1019-implicitly-add-the-module-suffix.md

kurtschelfthout:


dsyme:
Thank you, updates like this are invaluable

PR #154, opened 2016-12-08
Update FS-1028-Implement Async.StartImmediateAsTask.md

OnurGumus:
1. Changed double back ticks to single back ticks.
2. Added link for the discussion.


neoeinstein:
Thanks for the update.

PR #155, opened 2016-12-12
fix Markdown syntax

pocketberserker:


neoeinstein:
Thanks for the clean up!

PR #166, opened 2017-03-04
Add RFC for FSI Reference Model and extending #r

cartermp:


RFCs/FS-1027-fsi-references.md
```diff
@@ -0,0 +1,100 @@
...
+
+This extends the "language" of `#r` to support implementation and reference assemblies, NuGet packages, Paket dependencies, and potentially more.
+
+# Motivation
+[motivation]: #motivation
+
+Motivation for this change is twofold:
+
+1. There is a [strong desire](https://github.com/fsharp/fslang-suggestions/issues/542) to reference packages via `#r` instead of assemblies.  There has also been [experimental work](https://github.com/Microsoft/visualfsharp/pull/2483) to support Paket in this way.
+2. .NET Core uses both Reference Assemblies and Implementation Assemblies, which FSI must understand.  It does not understand this split today.
```
forki: I think the CSI needs to understand this as well. What's their plan?

RFCs/FS-1027-fsi-references.md
```diff
@@ -0,0 +1,100 @@
...
+2. Call into Paket to resolve dependencies.
+3. Generate a set of *Reference Assembly* and *Implementation Assembly* references in the ephemeral script, using `#r "path-to-ref-assembly"` and `#r "impl: path-to-impl-assembly""`
+4. Load the ephemeral script in the FSI session, thus giving FSI the correct references it needs to run in .NETCore
+4. Execute the above script which requires `Newtonsoft.Json`.
+
+## Errors
+
+FSI will have to display errors from the underlying assembly resolver.  For example, if someone specifies `#r paket: <nuget-command>` and Paket generates an error, FSI must display that.
+
+There is also another scenario to consider, which we will explicitly not handle.  For example, given the following:
```
forki: "this error needs to be raised by paket"

forki: And paket does that

RFCs/FS-1027-fsi-references.md
```diff
@@ -0,0 +1,100 @@
...
+
+printfn "Object: %A\n JSON: %s" foo json
+```
+
+When FSI encounters this script, it will:
+
+1. Create an ephemeral script.
+2. Call into Paket to resolve dependencies.
+3. Generate a set of *Reference Assembly* and *Implementation Assembly* references in the ephemeral script, using `#r "path-to-ref-assembly"` and `#r "impl: path-to-impl-assembly""`
+4. Load the ephemeral script in the FSI session, thus giving FSI the correct references it needs to run in .NETCore
```
forki: I thought #r-ref is the new one.

smoothdeveloper:
Question about Errors section (depends on two versions of transitive dependency), my intuition tells me that FSI is currently good enough to figure the binding redirects (as I never had binding redirect issue in a script), also paket will complete the resolution with at least the lowest version needed among the dependency set.

Is that something we plan will not work anymore?

The RFC is mixing the two syntaxes: `#r paket:"paket directives"` and `#r "paket: paket directives"`, I think we should now decide which one we want to adopt, the first one seem a bit more explicit as the second one look like a normal string and maybe confusing to people always expecting the string. Also `:` seems to be valid in filename on Linux, while not recommended this could be ambiguous.

RFCs/FS-1027-fsi-references.md
cartermp: Currently `#r-impl` or `#r "impl:` are the ones being considered, but that could just get hashed out in the issue.

RFCs/FS-1027-fsi-references.md
cartermp: @tmat could offer up more info, as I'm unaware of the plan for CSI.  It's a shared concern.  We're working with them here to make sure whatever we do makes sense.

cartermp:
@forki regarding your question:

> how do we hook these things in? Separate dll with marker attribute? Where would fsi look to find these dlls?

The package manager we use will give us locations that we need.  NuGet already does this, and I imagine that Paket offers up a similar facility when running on .NET Core (or defers to NuGet).

forki:
No what I meant is: how does fsi know which #r providers are available? In
the prototype we created a dll with the paket provider. If it's in same
path as fsi then it hooks in. The dll contains a type that's marked with
special attribute so that fsi can find it with reflection.

Am 07.03.2017 18:04 schrieb "Phillip Carter" <notifications@github.com>:

cartermp:
Moving discussion over to #167

PR #180, opened 2017-03-14
update load workflow

forki:


RFCs/FS-1027-fsi-references.md
```diff
@@ -114,25 +114,23 @@ Behavior should not be so different at runtime.  When executing code interactive
 
 When launching FSI with a script using one of the new `#r` references, FSI will do the following:
 
-1. Check if each dependency exists in the known place for the specified tool.  If it does, `#r` that dependency when executing FSI.
-2. If a dependency does not exist, search for the tool in a known/blessed area.  Error out if it can't be found.
-3. Call out to the tool to resolve dependencies.
-4. `#r` the resolved dependencies in an ephemeral script (or the script generated by Paket if using Paket).
-5. `#load` that script into the FSI session, thus giving it the references it needs to execute.
+1. Check which tools is registered for a given #r prefix.  Error out if no tool can't be found.
```
smoothdeveloper: Double negation not not not good

Error out if no tool can be found

RFCs/FS-1027-fsi-references.md
```diff
@@ -114,25 +114,23 @@ Behavior should not be so different at runtime.  When executing code interactive
...
+2. Let the tool check if each dependency exists in the known place for the specified tool or otherwise the the tool resolve all dependencies.
```
smoothdeveloper: The the

RFCs/FS-1027-fsi-references.md
```diff
@@ -164,6 +162,18 @@ This will generate an error because `System.Whatever` version `4.1.0` will alrea
 
 We may consider using `AssemblyLoadContext` in some clever way in the future, but it's very difficult to get right and well outside the scope of what we need to support for a .NET Core 2.0 timeframe.
 
+## Handler resolution
+
+FSI/Design time support will look at the following places in order:
+
+* if the script is a physical file, check current folder and browse all parent folders looking for .fsharp/fsi-extensions folder in each one
+* look into ~/.fsharp/fsi-extensions
```
dsyme: I wonder if this should be called ``fsx-extensions``.  ``FSI.EXE`` is one tool implementing the F# scripting model (i.e. the ``.fsx`` script extension). These extensions are relevant to all such tools.

RFCs/FS-1027-fsi-references.md
```diff
@@ -164,6 +162,18 @@ This will generate an error because `System.Whatever` version `4.1.0` will alrea
...
+* look into .fsharp/fsi-extensions folder next to fsi.exe
+
+gather all the distinct dll names, order of precedence favorising those in the same order shown above, and load them in the process if their assembly contains an arbitrary attribute (resolved by name rather than dependency on external library) and types marked with same attributes.
```
dsyme: We need more detail on the loading process here. 

* When you say ``gather all the distinct dll names`` do you look in immediate subdirectories?  I believe we should since each extension could be a whole tool of its own, with its own dependencies and .dl.config binding config?  

* Could you specify that for ``fsi,exe`` on .NET Framework these are loaded in-process using ``Assembly.LoadFrom``. 

* The same will apply to ``devenv.exe`` and other design-time tooling based on ``FSharp.Compiler.Service.dll`` such as anything based on ``FsAutoComplete.exe``.  We should specify whether this means the DLLs are loaded multiple times from different locations as you open and work with different scripts.

* It's possible that in the world of .NET Core this loading will look different.  I don't yet fully understand what the natural notion of an application extension is in the word of .NET Core and if there's a natural notion of a .NET Standard extension.  In any case we should note this and mention that we will need to specify whether extensions can be .NET Standard 1.6 DLLs and if so whether they will be loadable in both .NET Core and .NET Framework compilers

dsyme:
Added a comment, will merge and please submit a second PR

RFCs/FS-1027-fsi-references.md
smoothdeveloper: yes, sensible.

RFCs/FS-1027-fsi-references.md
smoothdeveloper: @dsyme 

for (1)

would it be ok to just look at the dlls in that folder? for paket we don't have (nor want) assembly reference to paket, the dll can load the assembly and/or work through reflection also.

it seems those entry points should be as lightweight as possible, and yes they can rely on having a subfolder in that same location, but does it need to be strong reference? isn't it better to not have that extension folder become a jungle of dll/folders?

We can also reintroduce a naming pattern, in current PR we have this: `"*DependencyManager*.dll"`, to not have to scan every single file, please advise and yes we can also browse subfolders (one level only?) if you like.

for (2) yes will do.

for (3):

>  We should specify whether this means the DLLs are loaded multiple times from different locations as you open and work with different scripts.

good point, that should be discussed, but the behaviour has somehow to match what we will have in FSI, this is one reason to browse folders up (like paket users do with .paket at root of repository), but maybe for tooling we should actually mandate a single or more restricted location instead?

for (4) ok.

smoothdeveloper:
@dsyme I made another PR: #181

PR #186, opened 2017-04-24
RFC FS-1032: Extend String module

saul:
Initial RFC proposal for https://github.com/fsharp/fslang-suggestions/issues/112


saul:
@cartermp as requested 😄

dsyme:
I like it too

PR #190, opened 2017-05-05
Correct file name for RFC FS-1033

mexx:
Rename FS-1032-extend-string-module.md to FS-1033-extend-string-modul…e.md


dsyme:
thanks!

PR #201, opened 2017-05-19
1034 - file size is not an issue

enricosada:
> The package is now relatively large (68MB on disk unzipped) due to the large number of variations, and may get bigger (E.g. embedded PDBs)

I dont think this is correct.

there are two things:

- nupkg compressed size: 7MB
- nupkg uncompressed size: 41.7MB (size and size on disk)

But there are also differnt things:

- bundles (VS/cli/Mono) ship the nupkg compressed, and put that in a directory, used as nuget local feed.

## local feed

The sdk and MS VS (check your `C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\` ) just add a directory of nuget package used as local feed. So compressed nupkg.

Cli on first use populate the local nuget cache, VS does that when a restore is needed (afaik)

My local VS nuget feed directory is 2.5GB. and FSharp.Core there is just 6,87MB (i added that for my offline scenario). There are 511 distinct package, with multiple version usually.

| Packages     | Size        | Count (by id) |
|-|-:|-:|
| Total            | 2500MB | 511 |
| Fsharp.Core |       7MB | 1 |

As a note, **i have there (not added by myself) multiple packages not owned by MS**, like `BundlerMinifier.Core`, `SQLite`, `SQLite.Native`, `Serilog.*` (5), `Remotion.Linq`, `Owin`, `StackExchange.Redis.StrongName`.
Who added these? dunno, but work for offline scenario

> * The package is not pre-installed with tooling (preventing some offline development scenarios)

because maybe can be done with the local feed. with package not owned by MS.

## local nuget cache

When the package is used, is restored in the local nuget cache. 

My directory in nuget cache for `Fsharp.Core` (`C:\Users\e.sada\.nuget\packages\FSharp.Core`) is 357MB and contains 13 version of Fsharp.Core (from 2.0.0, 3.0.2, ..., 4.1.17 ) 
My local nuget cache total size

| Packages     | Size        | Size (GB) | Count (by id) | Count (by version) |
|-|-:|-:|-:|-:|
| Total            | 40000MB| 40,0GB |1074 | dunno |
| Fsharp.Core |       357MB | 0,3GB | 1 | 13 |

### DIFF
--- a/RFCs/FS-1034-fsharp-core-package.md
+++ b/RFCs/FS-1034-fsharp-core-package.md
@@ -263,6 +263,13 @@ There are many others - searching github for ``nuget FSharp.Core`` in ``paket.de
   - Response:  No! See reasons above for  why this is not sensible - there are just too many scenarios where is it is just "way simpler and much easier" to instruct
     users to reference this package.
 
+## Notes
+
+* file size [more info and some stats](https://github.com/fsharp/fslang-design/pull/201), the 8MB nupkg is:
+  - the 6.5% of .net core sdk bundle, and 0.28% of VS local nupkg feed
+  - big (3-4 times) for a single library package
+
+
 ## Acknowledgements
 
 Thanks to Enrico Sada, Steffen Forkmann, Kevin Ransom, Phillip Carter for discussions leading up to the first draft of this RFC.


enricosada:
@dsyme is this a possibile way to make progress?
I'd like to discuss single point, to trim scope a bit, and hightligh the issues and scenario.

Some points i think are not in context.

enricosada:
As a note, while trimming obsolete tfm ihmo is ok, should be done separtaly next major (5.x) or minor (4.2), to avoid issues.

dsyme:
I think it's valid to question whether file size is really an issue and to base that on hard numbers.  But that analysis should be recorded as a valid question in the RFC with a record of the analysis, rather than just deleting the concern.

Separately we have to admit that the current FSharp.Core package was put together in a haphazard way (by myself included) and we should now work out if we can just ditch certain DLLs (e.g.  xamarinmacmobile) in favour of .NET Standard, for simplicity alone, but also for size.  I'm following up with @nosami and Xamarin about that.

The key statistics that matter to me are
* what does it add to an install of the .NET SDK
* how does it compare to other commonly used not-in-.NET-itself packages? e.g. choose some commonly used libraries

The comparison to a nuget cache is not so important for me. Could you get those statistics please? Thanks!

enricosada:
### packages

- `Automapper` => 500Kb
- `Nancy` => 1MB
- `Newtonsoft.Json` => 2MB
- `NUnit` => 2MB
- `SharpZipLib` => 2MB
- `SQLite` => 5MB

These support multiple target framework, usually 4+ 

*Important* Lot of these are splitted in multiple packages too, so usually are used in combination with at at least another (like `Serilog.*`, `Nancy.*`) so single packages are not always used alone, usually 2+ (one core + specifc to engine/tool) but some times more (see aspnet).

So for a library important as fsharp.core **and not splitted in multiple pkgs** 8MB is not that bad ihmo.

### .net core sdk bundle

An example `dotnet-1.1.1-sdk-win-x64.zip` is 122MB. 
Adding the existing `FSharp.Core` nupkg (8 MB) dunno if matter (because nuget cache is downloaded and initialized at first usage), but anyway adding inside as zip goes to 130MB (so `+ 6.5%`).


I'll update this later to move the info at bottom, instead of delete the point about file size

dsyme:
Thanks!

enricosada:
@dsyme added as notes at bottom, instead of replace.

dsyme:
@enricosada Thanks

PR #207, opened 2017-06-06
Create FSS-0001-option-of-string.md

isaacabraham:


dsyme:
@isaacabraham I've adjusted this to be a "draft RFC", since it's not really an approved RFC yet.  Could you create an fslang-suggestion issue for this and link to this doc as the draft RFC? thanks

dsyme:
Note draft RFCs don't really have numbers yet, though I guess we could use the issue number from the fslang-suggestions issue once created.

isaacabraham:
Sorry - yes, will do. It's my first commit :-)

dsyme:
> Sorry - yes, will do. It's my first commit :-)

@isaacabraham Even more... It's your first design change proposal :) Much,. much more work!

manofstick:
I think this would be better as `String.toOption`, as it seems to me to a bit to high level to be putting onto a very low level `Option`?

isaacabraham:
I'd be happy with that, too.

abelbraaksma:
Just like other APIs have their counterparts both implemented (`Seq.toArray` vs `Array.ofSeq`), I think you should do both here too: `String.toOption` and `Option.ofString`).

Since the Option type is meant for "do something when there is a value, do nothing otherwise" scenarios, I really think this should *not* take valid strings as `None`. And an empty or whitespace-only string is a valid string.

Consider `Option.ofObject`. Following the logic that an empty string should yield `None`, an empty object (i.e., `obj()`) should yield `None` too?

Since this is supposed to help prevent boilerplate or erroneous code by beginners, we should try to prevent people from writing:

```
input
|> Option.ofString
|> Option.bind (fun s -> if String.IsEmpty s then None else Some s)
|> Option.bind tryValidateInput

Some suggestions:

1. Perhaps add variants: `Option.ofString`, `Option.ofStringNonEmpty`, `Option.ofStringNonWhitespace` (the syntax help dropdown will make this clear at the spot)
2. Consider the inverse: `String.tryParse`, where an invalid string is the null-string (though admittedly, beginners may not find this function where they expect it, but it is in line with other core types and you can add String.parse which would throw an exception that is not a NullReferenceException, which helps in diagnostics)
3. Consider also adding whatever you add to `Option` to also add the inverse to `String` (as explained above)

PR #214, opened 2017-07-15
Update FS-1032: correct F# sdk properties

0x53A:
updates the F# Sdk properties to the actually used values.

TargetFSharpCorePackageVersion => FSharpCoreImplicitPackageVersion
DisableAutoFSharpCoreReference => DisableImplicitFSharpCoreReference
DisableAutoValueTupleReference => DisableImplicitSystemValueTupleReference

see https://github.com/Microsoft/visualfsharp/blob/master/src/fsharp/FSharp.Build/Microsoft.FSharp.NetSdk.targets#L60

cc @kevinransom

### DIFF
--- a/RFCs/FS-1032-fsharp-in-dotnet-sdk.md
+++ b/RFCs/FS-1032-fsharp-in-dotnet-sdk.md
@@ -190,18 +190,18 @@ This is valuable because System.ValueTuple is kind of confusing ...
 You need to reference it for net45,net46, net461, net462, net47 and coreapp1.+ and netstandard 1.6. You don't need to reference it for net471 and up or netcoreapp 2.+ or netstandard2.+.
 net47 is the most vexing, because the full desktop framework contains the type but the reference assemblies do not contain the type forward.
 
-__DisableAutoValueTupleReference__
+__DisableImplicitSystemValueTupleReference__
 To disable this, because ... reasons ...
-Use the DisableAutoValueTupleReference property [true or false]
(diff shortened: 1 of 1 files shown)


0x53A:
questions:
 
 * does TargetFSharpCorePackageIdentity still exist?
 * does TargetFSharpCoreVersion still exist?

I can't find them in the file above

dsyme:
@KevinRansom Any comments on the questions above?

KevinRansom:
Source is here:  https://github.com/Microsoft/visualfsharp/blob/master/src/fsharp/FSharp.Build/Microsoft.FSharp.NetSdk.targets#L71

Version controls are now named:
````
 FSharpCoreImplicitPackageVersion
 DisableImplicitFSharpCoreReference
 DisableImplicitSystemValueTupleReference
````
- FSharpCoreIdentity went away, because, we have standardized on FSharp.Core.nuget.
- FSharpCoreVersion is for Legacy FSharp projects.

Kevin

dsyme:
@KevinRansom Thx.  @0x53A Could you update with this info if you get the chance? thx

0x53A:
There is also a group [unresolved questions](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1032-fsharp-in-dotnet-sdk.md#unresolved-questions), which is a bit odd, since the feature is basically implemented.

Would you two mind just going over it once from top to bottom and adapting it to "really"? I think that is much easier from the inside than from me trying to reconstruct it from the outside.

PR #218, opened 2017-08-21
Create FS-1036-consistent-slicing.md

charlesroddie:


FS-1036-consistent-slicing.md
```diff
@@ -0,0 +1,77 @@
...
+
+Similarly `l.[i..]` should give:
+error when i < 0
+the last (l.Length-i) elements of l when 0 ≤ i ≤ l.Length
+[] when l.Length ≤ i
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None. It would be very hard for this to break any code.
```
dsyme: Drawback is that is is a breaking change if you are explicitly catching the exception.  (But that's not generally considered a significant breaking change)

Also potential perf impact )though I don't think that's a blocker)

charlesroddie: The list changes ( @syeerzy 's [PR](https://github.com/Microsoft/visualfsharp/pull/3475/fileslist) don't require any new checks, just replacing explicit `outOfRange()` with `[]` in the cases that are currently failing. So there shouldn't be a performance impact. I'd expect array and string changes to be very similar.

dsyme:
Super, thanks!

PR #235, opened 2017-11-26
Add FS-1042-add-seq-transpose.md

PatrickMcDonald:


RFCs/FS-1042-add-seq-transpose.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This is a standard operation on matrices (or lists of lists) which could be a useful addition to the core library.
+
+# Detailed design
+[design]: #detailed-design
+
+Given a collection of m collections of length n, transpose returns a collection of n collections of length m, where result.[i].[j] = source.[j].[i], for 0 <= i < n and 0 <= j < m
```
dsyme: Please add signatures.  Will you propose

    module List = 
        val transpose: 'T list list -> 'T list list

or

    module List = 
        val transpose: seq< #seq<'T> > -> 'T list list

or

    module List = 
        val transpose: seq< list 'T > -> 'T list list

The third matches ``List.concat`` and for that reason I suppose I'd recommend that one, but I could also see arguments for the first or second.

Note also that ``Seq.zip`` and friends don't fail on mismatched lengths.  Do you think ``Seq.transpose`` should be consistent with ``Seq.zip`` or ``List/Array.transpose``?

PatrickMcDonald: Will do. I implemented option 1 as it was simpler, but option 3, being consistent with concat, seems to be the way to go.  Option 2 does not have any lists in the input so seems least similar to existing functions.

I chose consistency with `Seq.zip` rather than `List/Array.transpose`. I guess otherwise you could have errors being thrown very late in an enumeration of a sequence.

PR #253, opened 2018-01-29
Remove x86/x64 from type loading spec, also add tools folder

dsyme:
cc @KevinRansom I think this captures the current implementation in visualfsharp master?


KevinRansom:
Yes, this is correct.

dsyme:
@KevinRansom Thanks!

PR #259, opened 2018-03-07
Create RFC FS-1048 - Extend ComputationExpression builder with Map

jwosty:


rmunn:
There's a mistake in this PR somewhere, I think. The filename says FS-1048 (adding a Map operation to computation expressions), but the content in this PR is currently FS-1047 (adding the `match!` operator to CEs).

jwosty:
@rmunn whoops, didn't notice that, thanks. I'll redo this tomorrow

PR #264, opened 2018-03-14
Add missing pipes

isaacabraham:
I've added a few missing | from the anonymous record braces. At least, I *think* that they are missing - please confirm, especially on line 309 (without them it looks identical to normal record copy-and-update syntax).


dsyme:
Thanks!

PR #267, opened 2018-03-15
Create FS-1049-nested-record-field--copy-and-update-expression

AviAvni:
initial work

### DIFF
--- a/RFCs/FS-1049-nested-record-field--copy-and-update-expression
+++ b/RFCs/FS-1049-nested-record-field--copy-and-update-expression
@@ -1,0 +1,56 @@
+# F# RFC FS-1049 - Nested Record Field Copy and Update Expression
+
+The design suggestion [Support first-class lensing / lenses](https://github.com/fsharp/fslang-suggestions/issues/379)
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
@AviAvni could you rename the file with .md extension?

Thanks!

AviAvni:
@smoothdeveloper thanks sorrt

dsyme:
Needs "open questions" and "alternatives" sections.  Especially cover these:

* What about indexers, e.g.

      let anotherPerson1 = { person with A.[3].N = person.A.[3].N + ", k.2" }

* What about similar features in the language that name fields, especially mutating property setters

      Person(A.C.N = 3, A.C.M = 4)

* What happens to cases where A.S is mentioned twice?  

       { person with A.S.N = person.A.S.N + ", k.2"; A.S.N = person.A.S.M + ", k.3"  }

I think I left some other comments in the PR thread too, please incorporate them here?

AviAvni:
@dsyme thanks for feedback the 3 questions are the thing you also wrote in the PR

PR #273, opened 2018-04-24
Create FST-1006-compiler-package-publish-vfsharp.md

cartermp:
This is a proposal that is intended to alleviate issues such as this: https://github.com/Microsoft/visualfsharp/issues/4782


ReedCopsey:
This seems like a great idea to me. I'd like to streamline this, and think the details w.r.t. ownership/etc should alleviate or prevent any significant concerns.

dsyme:
My thoughts:

* https://github.com/Microsoft/visualfsharp/issues/4782 is a non-issue, there was no strong reason given as to why an updated FCT was needed.  

* The "visualfsharp" repo doesn't yet do enough testing on Mono, e.g. against a whole matrix of Mono releases. That can be fixed, but it is work.  I am not comfortable with releasing an FCT without that testing.

* Traditionally, the community only rely on Microsoft to sign packages that are part of Microsoft's product surface area. FSharp.Core is part of the surface area, FCT is not.

PR #280, opened 2018-05-06
Update FS-1050-trygetvalue-map.md

forki:
We should not forget the map module

### DIFF
--- a/RFCs/FS-1050-trygetvalue-map.md
+++ b/RFCs/FS-1050-trygetvalue-map.md
@@ -1,6 +1,6 @@
 # F# RFC FS-1050 - Add TryGetValue on Map
 
-This proposes to add a `TryGetValue` method on the `FSharp.Collections.Map` type.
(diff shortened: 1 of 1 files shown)


RFCs/FS-1050-trygetvalue-map.md
```diff
@@ -1,6 +1,6 @@
 # F# RFC FS-1050 - Add TryGetValue on Map
 
-This proposes to add a `TryGetValue` method on the `FSharp.Collections.Map` type.
+This proposes to add a `TryGetValue` method on the `FSharp.Collections.Map` type and `tryGetValue` to the corresponding `Map` module.
```
dsyme: This is not made public in the draft implementation.  

What do you think? I'm not totally keen on making it public as it would be the first example of a public module-bound curried function taking a byref parameter.  In general I think of the module-bound functions as being in subset of F# API design that do not use "address bit fiddling".

I think it may be better to just have it on `Map` as a .NET-style instance member

dsyme:
closing as

PR #292, opened 2018-05-14
Move FS-1044 to FSharp.Core-4.4.3.0

dsyme:


dsyme:
This was implemented in FSharp.Core 4.4.3.0

PR #304, opened 2018-05-19
Fix (potential) typo in outref discussion

baronfel:
If I understand correctly, `outref` should be used here to signal a write-only handle.


dsyme:
Thanks!

PR #308, opened 2018-06-01
Update FS-1053-span.md

dsyme:
This updates the Span RFC with resolution of the outstanding design issues


dsyme:
@cartermp @jaredpar @TIHan please review. See also https://github.com/Microsoft/visualfsharp/pull/4888#issuecomment-393724789 for corresponding implementation update

PR #318, opened 2018-06-30
Added an expansion of the RFC abbreviation

ScottHutchinson:
For those who don't know, for instance, if it means "request for comments" or "request for changes".


dsyme:
I always think of it as "Request for Consideration" but "Request for Comments" is fine too

PR #319, opened 2018-07-05
Create FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md

7sharp9:


7sharp9:
@dsyme This is ready for review/merge etc

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
```diff
@@ -0,0 +1,122 @@
...
+    member x.Foo = 42
+    override x.Rotate _ = ()
+```
+
+This would result in the following warning:
+```
+[FS0060] Override implementations in augmentations are now deprecated. Override implementations should be given as part of the initial declaration of a type.
+```
+
+So the final part of this  proposal is to allow override implementations in provided types **only**, as it is not always possible to provide the implementation as part of the initial declaration of the type especially if you are only the consumer of the type provider.
```
cartermp: Nit: implementations in provided types for generative type providers only.

cartermp: Could you provide a concrete example of this being a limitation today? I think I understand why this is a limitation, but for the sake of documenting this relaxation of the deprecation for one specific case, an example would be helpful for others who go through and read this as the canonical source of why a change was made.

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
```diff
@@ -0,0 +1,122 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The only drawback is the complexity of the addition, but the addition makes provided types a lot more flexible especially as F# has no partial types, the olny alternative without this is inheriting the provided types and adding implementations there which is not ideal.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternative is to add features to the provided type by alternative means such as inheriting from the provided type or by augmenting another type which deferrs to the provided type internally.
```
cartermp: Why would inheriting from a provided type be a drawback compared with having intrinsic type extensions? I think we mostly agree that inheriting is icky unless it's the only way forward, but today this feels like such a scenario. Same questions for deferring to another type.

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
```diff
@@ -0,0 +1,122 @@
...
+[drawbacks]: #drawbacks
+
+The only drawback is the complexity of the addition, but the addition makes provided types a lot more flexible especially as F# has no partial types, the olny alternative without this is inheriting the provided types and adding implementations there which is not ideal.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternative is to add features to the provided type by alternative means such as inheriting from the provided type or by augmenting another type which deferrs to the provided type internally.  
+
+Another alternative would be to add partial classes to F# which is really what intrinsic type extensions are albeit limited to extensions in the same file.
```
cartermp: Nit: Probably worth adding that this would be an expensive alternative, with consequences that are far-reaching (e.g., changing how Go to Definition works in editor tooling).

7sharp9:
@cartermp I'll address those points tomorrow, its mainly due to the reason intrinsic type extensions were tried for provided types in the first place, providing code generation for Xamarin.iOS.  This framework as will many others requires types to look a certain way.

Types generated have to have attributes present and as the attributes don't propagate to the parent you have to repeat them or just not generate them on the abstract/base types and provide some kind of stringy helper for the user, I have an example of this on the iOS type provider.

Another reason is that many frameworks require an abstract type as the base class and certain methods overridden, _(again iOS required this)_ while its possible to do this by using inheritance it adds complexity, if you also take into account things like multiple constructor implementations are common then things end up using the record style multiple constructors, which again make the solution far less elegant.

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: done

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: done

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: done

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
```diff
@@ -79,19 +79,101 @@ This would result in the following warning:
...
+All of the duplicated constructor propagation can be avoided by allowing intrinsic type extensions  to occur on generative type providers.
+
+Method overloads are applied in the usual manner but if they could be applied as part of the 
+```fsharp
+    override x.ViewDidLoad() =
+        x.DoNotPressAction <- Some (fun _ -> x.View.BackgroundColor <- UIColor.Purple)
+        base.ViewDidLoad()
+```
+
+If this RFC was implemented the iOS type provider would look like this instead:
```
cartermp: Nit: is this supposed to say "Type provider" or "usage of the iOS type provider"?

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
```diff
@@ -0,0 +1,201 @@
...
+        base.DidReceiveMemoryWarning()
+        
+    override x.ViewDidLoad() =
+        x.DoNotPressAction <- Some (fun _ -> x.View.BackgroundColor <- UIColor.Purple)
+        base.ViewDidLoad()
+```
+
+This has several drawbacks that makes the user facing code more complex and unweidy than it should be with additions mentioned in the RFC.
+
+## Attributes have to be defined on outward facing type
```
cartermp: nit: can you use `###` for the header so that it's a sub-header of the limitations section?

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
```diff
@@ -0,0 +1,201 @@
...
+
+Many frameworks require the generation of attributes to be applied to generated type, Xamarin.iOS is such an example.  The `Register` attribute has to be applied to `ViewControllers` to supply the mapping between a user class and the storyboard name.  As attributes do not propagate with inheritance you cannot generate them on the base or abstract type.  The best option available to a type provider designer is to provide a helper to be generated that the consumer has to learn how to apply, of course this can be detailed in documentation but its an example of unnecessary work.  You can see this in the iOS UI provider above:
+
+```
+[<Register(container.myViewControllerBase.CustomClass)>]
+type ViewController(handle : IntPtr) =  
+```
+
+## Inheritance required
```
cartermp: Same here - `###`

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: the latter

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: fixed

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: fixed

RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md
7sharp9: updated text

vasily-kirichenko:
image

7sharp9:
Thanks @vasily-kirichenko 

Any more comments anyone?

dsyme:
I'll merge the RFC.  My main concern with this feature was that an implementation was pretty darn hard.

abelbraaksma:
For ease of browsing, it sits now here: https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1061-Allow-intrinsic-type-extensions-for-provided-types.md

PR #321, opened 2018-07-09
Update FS-1053-span.md

TIHan:


FSharp-4.5/FS-1053-span.md
```diff
@@ -264,21 +264,34 @@ let F1() =
...
+        if true then
+            &y
+        else
+            &z
+    ()
 ```
-This only applies when calling members, not arbitrary let-bound functions. See "Alternatives" below.
+`&y` will also give the same error.
+
+The rationale behind this is not just for soundness reasons, but for gaurantees in the optimizer that local values like this will not escape in such a manner.
```
dsyme: Typo `gaurantees`

FSharp-4.5/FS-1053-span.md
```diff
@@ -498,7 +511,7 @@ machinery to express read-only references.
 
 # Notes
 
-* Implicit address-of is not applied for parameters to let-bound functions.  For example:
+* Implicit address-of was intially implemented for member methods but was taken out due to scoping rationale. Currently it is **not implemented**. If it was done again, it wouldn't be applied for parameters to let-bound functions. For example:
```
dsyme: @TIHan Could you reword this a bit and place under `Alternatives` to capture the discussion we just had, e.g. that this is part of the corresponding C# design, and the example where the scope of the F# tmp is narrow, and that we will seek to re-instate this in F# 5.0 especially if API ergonomics make it necessary? thanks

TIHan: Yup will do.

FSharp-4.5/FS-1053-span.md
```diff
@@ -264,21 +264,34 @@ let F1() =
     c.InstanceProperty <- today.AddDays(2.0)
 ```
 
+#### Scoping for byrefs
 
-#### Implicit address-of when calling members with parameter type `inref<'T>` 
+The general idea is that a let-bound value cannot have its reference escape the scope from which it was declared, effectively its visibility. For example:
```
cartermp: What does "effectively its visibility" mean?

TIHan: The visibility scope of the value.

cartermp: Okay, I think this makes sense as-is.

PR #331, opened 2018-09-11
initial draft FS-1062-Async-WithCancellation

TheAngryByrd:


RFCs/FS-1062-Async-WithCancellation.md
```diff
@@ -0,0 +1,83 @@
+# F# RFC FS-1062 - (Add Async.WithCancellation)
+
+The design suggestion [FILL ME IN](https://github.com/fsharp/fslang-suggestions/issues/685) has been marked "approved in principle".
```
cartermp: `FILL ME IN` --> replace with the issue title

RFCs/FS-1062-Async-WithCancellation.md
```diff
@@ -0,0 +1,83 @@
...
+* [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/685)
+* [ ] Details: [under discussion](https://github.com/fsharp/fslang-design/issues/FILL-ME-IN)
+* [ ] Implementation: [In progress](https://github.com/Microsoft/visualfsharp/pull/FILL-ME-IN)
+
+
+# Summary
+[summary]: #summary
+
+
+This feature would add `Async.WithCancellation`.  This would allow using outside CancellationToken, such as the [RequestAborted](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpcontext.requestaborted?view=aspnetcore-2.1) from `HttpContext`.  The problem with [Async.Start](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/async.start-method-%5Bfsharp%5D?f=255&MSPPError=-2147217396) is it does not provide a way to return a value.
```
cartermp: Please add the second and third sentences under the Motivation section.

RFCs/FS-1062-Async-WithCancellation.md
```diff
@@ -0,0 +1,83 @@
...
+      tcs.TrySetException ex |> ignore }
+  Async.Start (a, cts.Token)
+  return! tcs.Task |> Async.AwaitTask }
+```
+
+Since all methods on `Async` are implemented as `static members` we should continue that pattern with the proposed signature. 
+
+`static member WithCancellation(a : Async<'a>, cancellationToken : CancellationToken)`
+
+with the same body as shown above.
```
cartermp: Could you add the full body implementation? Something like this:

```fsharp
type Async() = 
    ...
   static member WithCancellation(a: Async<'T>, cancellationToken: CancellationToken) = // body elided
````

File: RFCs/FS-1062-Async-WithCancellation.md [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I do no see any drawbacks to adding this feature.
```
cartermp: There is a drawback to adding more members like this, as it does move more towards having a paradox of choice w.r.t what to pick.

I personally feel that this is simply a net benefit to the library, but it's worth calling out that this is another way to do things, and it may not seem obvious to newcomers. But that negative is far outweighed by this solving a concrete problem.

cartermp: Additionally, could you phrase this without using the first person? `"I", "me", etc. shouldn't be in these RFCs.

RFCs/FS-1062-Async-WithCancellation.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Compatibility
+[compatibility]: #compatibility
+
+Please address all necessary compatibility questions:
+* Is this a breaking change? 
+
+No
+
+* What happens when previous versions of the F# compiler encounter this design addition as source code?
```
cartermp: These three questions are unanswered.

RFCs/FS-1062-Async-WithCancellation.md
```diff
@@ -0,0 +1,83 @@
...
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+* If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct?
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still TBD?
```
cartermp: I don't think there are any unresolved design questions here, so "None." should suffice.

RFCs/FS-1062-Async-WithCancellation.md
TheAngryByrd: If I don't have good answers should I just mark them as Unknown?

cartermp: I think we'll want these to be answered before a merge. At least for this first question, so long as this is a binary-compatible change (it is!) it should be fine. At least with FSharp.Core, the main concern is binary compatibility. Simple source changes like this won't break things in that way.

TheAngryByrd: Ok, this should be good for another pass.

RFCs/FS-1062-Async-WithCancellation.md
```diff
@@ -0,0 +1,107 @@
...
+[compatibility]: #compatibility
+
+Please address all necessary compatibility questions:
+* Is this a breaking change? 
+
+No
+
+* What happens when previous versions of the F# compiler encounter this design addition as source code?
+
+It will not compile.
```
cartermp: In this case, I would expect it to compile, since the change is source and binary compatible (as far as I can tell.)

cartermp: Actually, in this case, I'd say that "N/A" is the appropriate reply for this question, since it's more for new language constructs rather than FSharp.Core constructs.

PR #334, opened 2018-09-24
Add let! ... and! ... RFC

TD5:
Currently just a first-draft


TD5:
Discussion: https://github.com/fsharp/fslang-design/issues/335

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+        }
+    if quux > 6
+    then
+        return quux
+    else
+        return 5
+}
+```
+
+?
```
cartermp: What is the significance of `?` here?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+
+In order to keep things simple, then, we keep the canonical form introduced earlier, which forces precisely one `return` after a `let! ... and! ...`. However, we can still express alternative applicatives when using the `let! ... and! ...` syntax, we just need to use the same trick as with additional `let!`s and leave the scope of the canonical applicative syntax and therefore leave the additional constraints it places upon us:
```
cartermp: Upon reading this section, I think it needs to be shuffled around a bit:

1. Explain that you can use `yield`
1. Explain the restrictions (i.e., must follow a canonical form)
1. Explain why this is required

Additionally, I suggest renaming the section to either `Using yield` or `Using monoids`.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+            bar),
+        baz)
+)
+```
+
+*N.B.* this syntax forces the writer to be explicit about how many times `Apply` should be called, and with which arguments, for each call to `Combine`. Notice also how the right-hand sides are still repeated for each alternative case in order to keep the occurrence of potential side-effects from evaluating them predictable, and also occur before the pattern matching _each time_ a new alternative case is explored.
+
+If this syntax feels a bit heavy, remember that the `yield` keyword is not required, and the new syntax can be mixed with other styles (e.g. a custom `<|>` alternation operator) to strike an appropriate balance as each situation requires.
+
+## Using
```
cartermp: Can you also clarify that the same retrictions for `let! .. and!` apply here?

cartermp: That is, assuming they do (which I'm pretty sure they would :))

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+
+Or via `Apply`:
+
+```fsharp
+ce.Apply(
+    ce.Return(fun x -> x + 1),
+    foo)
+```
+
+This is because the operation is really equivalent to a `Map`, something which can be implemented in terms of `Return` and either `Bind` or `Apply`. We mentioned that these functions were in some sense more powerful than a plain functor's `Map`, and we are seeing an example of that here.
```
cartermp: Can you clarify "more powerful" here?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+                )),
+            foo), 
+        bar),
+    baz)
+```
+
+## Ambiguities surrounding a `let! .. return ...`
+[singlelet]: #singlelet
+
+Some CEs could be validly desugared in multiple ways, depending on which methods are implemented on a builder (assuming the implementations follow the standard laws relating these functions).
```
cartermp: Please omit "validly" from this:

"Some CEs could be validly desugared in multiple ways[...]"

It reads cleaner and basically says the same thing.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+```
+
+## What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+
+Since the syntax is desugared into a method call on the builder object, after compilation usages of this feature will be usable with previous compiler versions.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Is `anduse!` an acceptable keyword for when `and!` must also imply a call to `Using`?
```
cartermp: Just a comment here, any alternative syntax you think of can be listed under Alternative Designs, just under a subsection entitled "Syntax".

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+
+There are various ways of desugaring `let! ... return ...` via one of `Map`, `Apply` or `Bind`. [The above](#singlelet) assumes that we are aiming for backward compatibility and hence doesn't consider desugaring to `Apply` at all. Is this the best choice? Alternatives include:
+
+* Using `Apply` in preference to `Bind` for `let! ... return` whenever `Apply` is defined (on the basis that that any reasonable `Apply` implementation will be functionally equivalent, but more efficient than the corresponding `Bind`)
+* Subsuming [the RFC for desugaring this to `Map`](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1048-ce-builder-map.md) and defining a hierarchy between `Map`, `Apply` and `Bind` and some attributes for those methods to allow the creators of builders to opt-in to "optimisations" that pick the least powerful (and hence hopefully most efficient) desugaring.
```
cartermp: Another mention of powerful here that may need clarification. Perhaps doing so with a short glossary of terms of the top is better?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+)
+```
+
+*N.B.* this syntax forces the writer to be explicit about how many times `Apply` should be called, and with which arguments, for each call to `Combine`. Notice also how the right-hand sides are still repeated for each alternative case in order to keep the occurrence of potential side-effects from evaluating them predictable, and also occur before the pattern matching _each time_ a new alternative case is explored.
+
+If this syntax feels a bit heavy, remember that the `yield` keyword is not required, and the new syntax can be mixed with other styles (e.g. a custom `<|>` alternation operator) to strike an appropriate balance as each situation requires.
+
+## Using
+
+Just as monads support `Using` via `use!`, applicatives supports it via `use! ... anduse! ...`. Each binding can be either `and!` or `anduse!` (unless it is the first, in which case it must be either `let!` or `use!`), i.e. you can mix-and-match to describe which bindings should be covered by a call to `Using`:
```
cartermp: `anduse!` is interesting, since it does sort of break the visual symmetry you can get with `let!`, `use!`, and `and!`. I'm not opposed to it, mostly because I don't have a good alternative, but it does feel a bit odd. Limitation of the English language, I suppose.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+
+> Any expression built from the Applicative combinators can be transformed to a canonical form in which a single pure function is "applied" to the effectful parts in depth-first order:  
+`pure f <*> arg1 <*> ... <*> argN`  
+This canonical form captures the essence of Applicative programming: computations have a fixed structure, given by the pure function, and a sequence of subcomputations, given by the effectful arguments.
+
+In our case, the expression to the right of `return` (i.e. `pure`) becomes the body of a lambda, whose parameters are introduced by the `let! ... and! ...` preceding it. This leads to a very straightforward desugaring of the syntax, which therefore ensures both using and understanding the feature is only as complex as is inherently required by the abstraction.
+
+Despite requiring the canonical form, there are still many ways to build more complex and useful expressions from this syntax. The rest of this section aims to give a tour around these various features.
+
```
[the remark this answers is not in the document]
TD5: This is something I've largely delayed looking into too deeply so far. iirc, in my current implementation, not having a `return` after a `let! ... and! ...` gives (or could easily give) a custom error message. The ranges do track the individual bindings, so the squigglies aren't always going to just cover the whole CE, although this also probably wants a bit of refinement, or at least a check to make sure it's not misleading.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: These were arrows, but it looks like I accidentally corrupted them at some point. I'll fix that.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: I totally agree. I've asked for alternative suggestions for this keyword at the bottom of the RFC and the discussion. It would be nice to have something a bit more tidy here.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+# Summary
+[summary]: #summary
+
+Extend computation expressions to support applicative functors via a new `let! ... and! ... return ...` syntax.
+
+# Motivation
+[motivation]: #motivation
+
+Applicative functors (or just "applicatives", for short) have been growing in popularity as a way to build applications and model certain domains over the last decade or so, since McBride and Paterson published [Applicative Programming with Effects](http://www.staff.city.ac.uk/~ross/papers/Applicative.html). Applicatives are now reaching a level of popularity within the community that supporting them with a convenient and readable syntax, as we do for monads, makes sense.
```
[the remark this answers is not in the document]
TD5: I've added a little example just below to give a flavour of what is to come. Do you think that is enough?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+
+[Applicative functors](https://en.wikipedia.org/wiki/Applicative_functor) sit, in terms of power, somewhere between [functors](https://en.wikipedia.org/wiki/Functor#Computer_implementations) (i.e. types which support `Map`), and [monads](https://en.wikipedia.org/wiki/Monad_(functional_programming)) (i.e. types which support `Bind`).
+
+If we consider `Bind : M<'T> * ('T -> M<'U>) -> M<'U>`, we can see that the second element of the input is a function that requires a value to create the resulting "wrapped value".
+
+In contrast, `Apply : M<'T -> 'U> * M<'T> -> M<'U>` only needs a wrapped function, which is something we have whilst building our computation.
+
+So, importantly, applicatives allow us the power to use functions which are "wrapped up" inside a functor, but [preserve our ability to analyse the structure of the computation](https://paolocapriotti.com/assets/applicative.pdf) - and hence allow for the introduction of optimisations or alternative interpretations - without any values present. This is a critical distinction which can have a huge impact on performance, and indeed on what is possible to construct at all, so has very tangible implications.
+
```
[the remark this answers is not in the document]
TD5: I've touched this in light of your comment, but I think I will revisit this again at some point.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,432 @@
...
+    baz)
+```
+
+To be accepted as an applicative computation expression (CE), the CE must be of the form `let! ... and! ... return ...` (with one or more `and!`s). This means, for example, no `let!`s after an `and!` in the same CE, no normal `let`s in between the `let!` and `and!`s, and no usages of `yield` in place of `return`. Similarly, `do!` and `match!` are not valid in an applicative builder, although we allow `use!` and related keywords, which will be covered later.
+
+This may sound very constrained, but it is for good reason. The structure imposed by this rule forces the CE to be in a canonical form ([McBride & Paterson](http://www.staff.city.ac.uk/~ross/papers/Applicative.html)):
+
+> Any expression built from the Applicative combinators can be transformed to a canonical form in which a single pure function is "applied" to the effectful parts in depth-first order:  
+`pure f <*> arg1 <*> ... <*> argN`  
+This canonical form captures the essence of Applicative programming: computations have a fixed structure, given by the pure function, and a sequence of subcomputations, given by the effectful arguments.
```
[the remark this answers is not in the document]
TD5: I've just added this below:

> Similarly, the canonical form of `let! ... and! ... return ...` in F# makes should make it clear that what we are really doing it calling the function given to `return` with the arguments introduced by `let! ... and! ...`, but in a special context determined by the CE builder.

TD5:
I've modified the section on managing resources. My previous proposal was wrong because it desugared `and!` and `anduse!` to include a call to `builder.Using(...)`. This is wrong because, inside an applicative, using needs a different type signature (the previous one required a type signature which implied a monad, now we only need on imply a functor - hence it should be compatible with the [the RFC for desugaring this to `Map`](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1048-ce-builder-map.md) which doesn't seem to address this).

```fsharp
Using : 'T * ('T -> M<'U>) -> M<'U> when 'U :> IDisposable
```

in comparison to

```fsharp
MapUsing : 'T * ('T -> 'U) -> 'U when 'U :> IDisposable
```

fwiw, I haven't totally proved this new change out, but I thought I should mention is before someone else goes down a rabbit hole of wondering what on earth I was suggesting there.

TD5:
@cartermp, I've made some changes off the back of your feedback. Do you have any more suggestions or comments?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,746 @@
...
+[summary]: #summary
+
+Extend computation expressions to support applicative functors via a new `let! ... and! ... return ...` syntax.
+
+# Motivation
+[motivation]: #motivation
+
+Applicative functors (or just "applicatives", for short) have been growing in popularity as a way to build applications and model certain domains over the last decade or so, since McBride and Paterson published [Applicative Programming with Effects](http://www.staff.city.ac.uk/~ross/papers/Applicative.html). Applicatives are now reaching a level of popularity within the community that supporting them with a convenient and readable syntax, as we do for monads, makes sense.
+
+With an applicative computation expressions, we can write more computations than before (there are more contexts which meet the requirements for applicative computation expressions than the existing monadic ones), and we can write more efficient computations (the requirements of applicatives rule out needing to support some potentially expensive operations).
```
alisterw: > With an applicative computation expressions

Minor typo

TD5: Good spot! Thanks @alisterw!

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,750 @@
...
+
+# Summary
+[summary]: #summary
+
+Extend computation expressions to support applicative functors via a new `let! ... and! ... return ...` syntax.
+
+# Motivation
+[motivation]: #motivation
+
+Applicative functors (or just "applicatives", for short) have been growing in popularity as a way to build applications and model certain domains over the last decade or so, since McBride and Paterson published [Applicative Programming with Effects](http://www.staff.city.ac.uk/~ross/papers/Applicative.html). Applicatives are now reaching a level of popularity within the community that supporting them with a convenient and readable syntax, as we do for monads, makes sense.
```
panesofglass: I would argue it's not specifically applicatives but something along the lines of "at the same time," e.g. parallel, concurrent, etc. that we should push as the focal point. Applicatives are the category theory term, but computation expressions are not merely about category theory. I think more people will understand the desire for, e.g. in `async` computations, to await multiple, unrelated computations at the same time rather than _applicatives_ in particular. That said, references to applicatives are certainly useful as a means of showing why the implementation may follow a certain path.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,750 @@
...
+For example, [Pauan points out](https://github.com/fsharp/fslang-suggestions/issues/579#issuecomment-310799948) that we can write a computation expression for `Observable`s that [avoids unnecessary resubscriptions](https://github.com/fsharp/fslang-suggestions/issues/579#issuecomment-310854419):
+
+```fsharp
+// Outputs a + b, which is recomputed every time foo or bar outputs a new value,
+// avoiding any unnecessary resubscriptions
+observable {
+  let! a = foo
+  and! b = bar
+  return a + b
+}
```
panesofglass: For computation expressions, this is a very interesting case. What is meant by `let! ... and! ...`? This looks, on first glance, like an example of `Observable.zip`. However, what's to stop someone from implementing this as a pre-defined `Observable.Join`? I think both are valid implementations for a computation expression, depending on use case.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,750 @@
...
+```
+
+## Why Applicatives?
+
+[Applicative functors](https://en.wikipedia.org/wiki/Applicative_functor) sit, on the spectrum of flexibility vs. predictability, somewhere between [functors](https://en.wikipedia.org/wiki/Functor#Computer_implementations) (i.e. types which support `Map`), and [monads](https://en.wikipedia.org/wiki/Monad_(functional_programming)) (i.e. types which support `Bind`, which currently underpin computation expressions).
+
+If we consider `Bind : M<'T> * ('T -> M<'U>) -> M<'U>`, we can see that the second element of the input is a function that requires a value to create the resulting "wrapped value". This means the argument to `Bind` has the power to completely change the context of the result based on the value seen (e.g. to create and destroy `Observable` subscriptions), but it also means that the expression builder can predict much less about what the given function will decide to do, and hence has fewer outcomes that it can rule out and potentially optimise away.
```
panesofglass: It's worth noting that while the documentation specifies this `M` remains consistent throughout, you can write valid computation expressions (especially type extensions) that replace an input `M` with another type, e.g. type extensions for `AsyncBuilder` that allow binding `Task<'T>` or `Task`.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,750 @@
...
+
+In contrast, `Apply : M<'T -> 'U> * M<'T> -> M<'U>` only needs a wrapped function, which is something we have whilst building our computation and not something that can be controlled by the values at come later. This removes some flexibility to drastically alter the shape of the context in response to values seen later, but means that the computation expression builder now knows much more about what can or cannot happen after construction, and hence can make intelligent decisions off the back of that (e.g. to avoid unsubscribing only to immediately resubscribe, or to perhaps run two operations in parallel because it knows there can be no dependencies between them).
```
panesofglass: `Apply` feels to me to specific for the nature of computation expressions. I lean more to the `Merge` + `Map` proposed in [The F# Computation Expression Zoo](http://tomasp.net/academic/papers/computation-zoo/). I think it is at least worth adding a section discussing this option and showing expansions and use cases as to why `Apply` is the preferred option here.

panesofglass: NOTE: I wanted to note this as I suspect `Map` and `Merge` are slightly more general and thus more in the spirit of F# computation expressions.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: Yes, I did have this sort of thing in mind at some point and obviously failed to put it into the RFC at any point. I'll revisit this, because I think it's an important aspect.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: Interesting thought. I'd been working on the basis that putting in anything but a consistent `M` was going off-piste somewhat, so I planned to skirt over it. Perhaps it is worth covering that after all.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: It was my understanding that `Map` and `Merge` offer precisely the power of `Apply` and `Return`. However, `Merge` introduces the notion of tuples, which I find inelegant, but more importantly, `Apply` is the standard formulation that one would see in the wider literature.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: This isn't something that I know much about or have given a vast amount of thought to, so I could be wrong on this: Nothing special here apart from there being a convenient syntax for doing something like `Observable.zip` but for any number of `Observable`s. Note also how the new syntax forces names to be given to each parameter right next to where the argument is supplied, making it harder to accidentally swap `Observable`s being passed in which are of the same type.

I don't know what you would necessarily mean by `Observable.Join`.

I think the point here is that I perhaps wrote a slightly misleading description in the RFC. It's not that something sensible is inexpressible now (e.g. via `Observable.zip`), but more that if you wanted to use some nice computation expression syntax for it, you'd need to pay an annoying performance cost.

TD5: I'll look to make that clearer in the RFC.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: fwiw, I want to be careful to avoid baking in the `async` as the canonical example as the intuition for this - the feature is much more general than that.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
panesofglass: I think the big challenge with computation expressions is remembering that `Bind` and `Return` are not exactly equal to the `bind` and `return` of monads. It's common to fall into the trap of thinking of them in this way. I don't think it a _wrong_ interpretation; it's just that CEs are _not_ monad builders or applicative builders, though you can encode such constructs within them.

I don't think `Apply` is a bad direction; however, I think it worth expressing the expansions side-by-side and thinking about the tradeoffs. For example, if `Apply` and `Return` offer a wider range of expressions within CEs, then we should use that. If `Map` and `Merge` offer a wider expression, then we should use that. Category should only enter the picture to ensure that such constructs _can_ be encoded. They should not, imo, be the sole purpose for expanding the builder syntax. I think this is in keeping with the spirit of how they have been implemented to date.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
panesofglass: Another example would be allowing `Bind` (or `Apply` in this case`) against a `null` or `Nullable<'T>` in an `OptionBuilder`.

panesofglass: I highlighted some of this flexibility in https://github.com/panesofglass/computation-expressions-workshop

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
voronoipotato: Applicatives culturally did not directly come from category theory, they came out of programming and are merely associated with category theory through functors. That's why I put "proper" in quotes, because *technically* there is no such thing as an applicative in category theory. It's much more useful to us programmers and less interesting to pure mathematicians so it gets a special name for us. I think breaking it out into map and merge add indirection in a way that makes it less accessible, additionally it's more work. Map is already being considered in a different RFC but considering merge. Merge lets us take a tuple where each term is the same container type, and return a container type that contains a tuple. Apply lets me sidestep the tuple dance and stay in whatever container type I'm in. I can see the value of using a monoidal conceptual model in Tomas' terms "affect what (applicative) computations are evaluated as part of the computation.". However I think defining an intuitive syntax and good UX be a much more complex and involved process, that's why I think merge and map computation expressions should be its own RFC. It looks like Tomas has already put some effort into designing such a thing, but has not given any good applications outside of formlets. Your standard applicative functor is much simpler syntactically, I would hate for merge/map to become something that prevents us from ever actually getting Apply, something which has many more uses and is easier to implement and test.


http://tryjoinads.org/index.html?computations/applicative.html
Apply 
``` fsharp
val apply : F<'T1 -> 'T2> -> F<'T1> -> F<'T2>

val merge:  F<'T1> * F<'T2> -> F<'T1 * 'T2>
```

panesofglass: That is a good explanation. Thanks!

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: I've added an example of parallelism because, despite my fears, it is a very useful example of the power of applicatives. I've tried to emphasise the independence of arguments as part of that.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: I've subtly changed the wording around this in an attempt to clarify that it wasn't impossible to implement applicatives before, but that this is just convenient syntax for them.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: The [MSDN docs](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/computation-expressions#creating-a-new-type-of-computation-expression) talk about "typical signatures" of what is being desugared. I suppose that is a bit different from when I am coming from in this RFC. I don't think basing this change in the established literature and theory is a bad thing, but mentioning that you _can_ use other signatures I think probably is important, as you say.

TD5: I've added a small section dedicated to clarifying that other method signatures are possible.

TD5:
@cartermp / @dsyme: I am reasonably happy with the state of the RFC now. I suspect I'll make some provisional executive decisions regarding the open questions just to keep things moving. From your perspectives, does this RFC meet the requirements to be accepted?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+[summary]: #summary
+
+Extend computation expressions to support applicative functors via a new `let! ... and! ... return ...` syntax.
+
+# Motivation
+[motivation]: #motivation
+
+Applicative functors (or just "applicatives", for short) have been growing in popularity as a way to build applications and model certain domains over the last decade or so, since McBride and Paterson published [Applicative Programming with Effects](http://www.staff.city.ac.uk/~ross/papers/Applicative.html). Applicatives are now reaching a level of popularity within the community that supporting them with a convenient and readable syntax, as we do for monads, makes sense.
+
+With applicative computation expressions, we can write more computations with this convenient syntax than before (there are more contexts which meet the requirements for applicative computation expressions than the existing monadic ones), and we can write more efficient computations (the requirements of applicatives rule out needing to support some potentially expensive operations).
```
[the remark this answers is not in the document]
cartermp: This example should probably exist in the Summary section.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+// Outputs a + b + c, which is recomputed every time foo or bar outputs a new value,
+// avoiding any unnecessary resubscriptions
+observable {
+  let! a = foo
+  and! b = bar
+  and! c = baz
+  return a + b + c
+}
+```
+
```
cartermp: Can you add a sentence or two that explains what you would need to do without such a syntax? Just has to be high-level, but that helps cement the motivation.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+In contrast, `Apply : M<'T -> 'U> * M<'T> -> M<'U>` only needs a wrapped function, which is something we have whilst building our computation and not something that can be controlled by the values at come later. This removes some flexibility to drastically alter the shape of the context in response to values seen later, but means that the computation expression builder now knows much more about what can or cannot happen after construction, and hence can make intelligent decisions off the back of that (e.g. to avoid unsubscribing only to immediately resubscribe, or to perhaps run two operations in parallel because it knows there can be no dependencies between them).
```
cartermp: "by the values at come later" ---> "by the values **that** come later"

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+
+So, importantly, applicatives allow us the power to use functions which are "wrapped up" inside a functor, but [preserve our ability to analyse the structure of the computation](https://paolocapriotti.com/assets/applicative.pdf). This is a critical distinction which can have a huge impact on performance, and indeed on what is possible to construct at all, so has very tangible implications.
```
cartermp: It would be helpful to show an attempt to do some parallel processing _without_ applicatives, then describe why they are limited when compared with applicatives. This would help cement the _why_ in this section.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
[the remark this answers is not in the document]
cartermp: I think what you've touched up is sufficient.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+// Outputs a + b, which is recomputed every time foo or bar outputs a new value,
+// avoiding any unnecessary resubscriptions
+observable {
+  let! a = foo
+  and! b = bar
+  return a + b
+}
+```
+
+[McBride & Paterson's paper](http://www.staff.city.ac.uk/~ross/papers/Applicative.html) introduces a type very much like F#'s `Result<'T,'TError>` which can be used to stitch together functions and values which might fail, but conveniently accumulating up all of the errors which can then be helpfully presented at once, as opposed to immediately presenting the first error. This allows you to take [Scott Wlaschin's Railway Oriented Programming](https://fsharpforfunandprofit.com/rop/) to the next level by not just bailing out when things go wrong, but automatically collecting up a useful description of all of the different issues that occurred that resulted in the failure.
```
cartermp: Nit: "Collect up" ---> "collect"

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+// be, etc.) and the information needed to generate its help text
+opt {
+  let! username = Opt("username", (Some ""), Some)
+  and! fullname = Opt("fullname", None, Some)
+  and! id = Opt("id", None, readInt)
+  return User(username, fullname, id)
+}
+```
+
+With all of these examples, we can nest the applicative computation expressions inside other computation expressions to build larger and larger descriptions that cleanly separate the pure computation from its context.
```
cartermp: Nit: "larger and larger" --> "larger"

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+                            x + y + z
+                        )
+                    )
+                )),
+            foo),
+        bar),
+    baz)
+```
+
+## Rules on keywords in applicative computation expressions
```
cartermp: Prior to this section there should be a catalog of the new keywords and what they significance is. You allude to this a bit in this section and with code samples prior to this, but a detailed list of the new keywords will be important before you describe their rules.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+The `let! ... and! ...` form is replaced entirely by its resource-tracking equivalent:
+```fsharp
+ce {
+    use!    x = foo ✔️
+    anduse! y = bar ✔️
+    anduse! z = qux ✔️
+    return x + y + z
+ }
+```
+
```
cartermp: Is there  a valid use with combined with `return!`? If so, let's add an example.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,769 @@
...
+To be accepted as an applicative computation expression (CE), the CE must be of the form `let! ... and! ... return ...`:
+
+* There must be at least one `and!`s after the `let!`, but there is no hard upper-limit on the number of `and!`s
+* No `let!`s may appear after an `and!` in the same applicative CE
+* No normal `let`s in between the `let!` and `and!`s
+* No usage of `yield` in place of `return`
+* `do!`, `match!` and other CE keywords are also not valid in an applicative CE
+* `use!` and `anduse!` may replace `let!` and `and!`, respectively, to indicate a resource that needs to be managed by the CE builder
+
```
cartermp: If `return!` is a part of this equation, let's also call it out in the rules.

cartermp: Additionally, though you had mentioned that you didn't account for errors yet, I think this RFC is at a place where this can be done. I think we should have a table of error messages that will be added when combining things incorrectly, and for the negative examples below, include what is emitted. In short, for every invalid combination we provide a code sample for, we should have an appropriate error message that we say will be emitted.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
cartermp: Thinking about this, I the answer is probably no.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,832 @@
...
+In our case, the expression to the right of `return` (i.e. `pure`) becomes the body of a lambda, whose parameters are introduced by the `let! ... and! ...` preceding it.
+
+Similarly, the canonical form of `let! ... and! ... return ...` in F# makes should make it clear that what we are really doing it calling the function given to `return` with the arguments introduced by `let! ... and! ...`, but in a special context determined by the CE builder.
+
+Despite requiring the canonical form, there are still many ways to build more complex and useful expressions from this syntax. The rest of this section will detail these.
+
+## Desugaring
+
+Computation expressions are provided meaning via [translation to method calls on a builder class](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/computation-expressions#creating-a-new-type-of-computation-expression).
```
alisterw: Is it worth explicitly mentioning the expected types of the builder methods (as in the table in the [existing documentation for CEs)](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/computation-expressions#creating-a-new-type-of-computation-expression), either here or elsewhere, or do you think that's unnecessary?

(The type of Apply is given in the "Why applicatives?" section, but that seems to be almost in passing.)

TD5: Good point! That has been sat in my notes for weeks and I'd just kind of presumed it was in the RFC somewhere already.

Thanks!

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,890 @@
...
+```
+
+## What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+
+Since the syntax is desugared into a method call on the builder object, after compilation usages of this feature will be usable with previous compiler versions.
+
+# Unresolved Questions
+[unresolved]: #unresolved-questions
+
+Is `anduse!` an acceptable keyword for when `and!` must also imply a call to `Using`? Are there sensible alternatives?
```
dsyme: I'm guess I'm ok with `anduse!` all things considered, though perhaps we need a really compelling example where multiple resources are bound in a way that needs an imperative disposal.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,890 @@
...
+
+There are various ways of desugaring `let! ... return ...` via one of `Map`, `Apply` or `Bind`. This RFC currently assumes that we are aiming for backward compatibility and hence doesn't consider desugaring to `Apply` at all. Is this the best choice? Alternatives include:
+
+* Using `Apply` in preference to `Bind` for `let! ... return` whenever `Apply` is defined (on the basis that that any reasonable `Apply` implementation will be functionally equivalent, but more efficient than the corresponding `Bind`)
+* Subsuming [the RFC for desugaring this to `Map`](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1048-ce-builder-map.md) and defining a hierarchy between `Map`, `Apply` and `Bind` and some attributes for those methods to allow the creators of builders to opt-in to "optimisations" that pick the least general (and hence hopefully most efficient) desugaring (i.e. pick `Map` if it is defined, else `Apply`, else `Bind`).
```
dsyme: I think it would be fantastic if this RFC could also address [FS 1048](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1048-ce-builder-map.md).  It does feel like the second choice here is the best one, i.e.  "pick `Map` if it is defined, else `Apply`, else `Bind`"

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,890 @@
...
+* Using `Apply` in place of `Bind` only in those instances where `Bind` is not defined (no existing code should break, but this goes somewhat to the contrary of the previous proposal, which we may want to consider in the future)
+
+What alternative names are there for `MapUsing : 'T * ('T -> 'U) -> 'U when 'U :> IDisposable`? The prefix "Map" is wrong because we're not inside some context given by a type here. If anything, I'd say this should be called `Using` and the existing thing should be `BindUsing` but since we can't realistically change that, we'll need a new name. Perhaps `ApplyUsing`, but this method isn't intrinsically linked to `Apply` in that we could equally use it for `Bind` if things desugared differently.
```
dsyme: I think I prefer `ApplyUsing`.  Agreed that `Using` should have been called `BindUsing`.

dsyme:
I'm ok to pull in this RFC, I left a few comments to resolve the unresolbed issues.  Fantastic work here

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,955 @@
...
+    foo)
+```
+
+This is because the operation is really equivalent to a `Map`, something which can be implemented in terms of `Return` and either `Bind` or `Apply`. It is in this sense that these functions are more general than a plain functor's `Map`.
+
+In order to avoid breaking backwards compatibility, the default resolution is to desugar via `Bind`, _failing if it is not defined on the builder_ (even though, conceptually, it could be implemented via `Apply`). This is consistent with in previous F# versions. [Later work on supporting `Map`](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1048-ce-builder-map.md) can then make the choice about how to resolve this in a way which works with that in mind too.
+
+### Run and Delay support
+
+When a computation expression builder implements a `Run` or `Delay` method (or both), the desugared computation expression in wrapped in further calls corresponding to what is defined on the builder. This is true for both monadic and applicative computation expressions.
```
alisterw: "in wrapped" -> "is wrapped"

TD5: Wow, that was fast!

dsyme:
@cartermp merge when you're ok with it? thanks :)

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
cartermp: Actually I think this is fine without that.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
cartermp: It's probably a bit much to show the previous-world example.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
cartermp: You've done some of that here which is good.

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -0,0 +1,957 @@
...
+Pattern matching in conjunction with `use!` or `anduse!`:
+```fsharp
+ce {
+    use! (_,x) = foo
+    anduse! y  = bar
+    return x + y
+ }
+// Example.fsx(2,10): error FS3244: Pattern matching is not allowed on the left-hand side of the equals. 'use! ... anduse! ...' bindings must be of the form 'use! <var> = <expr>' or 'anduse! <var> = <expr>'.
+```
+
```
cartermp: Can you add examples of using the unsupported keywords?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
TD5: Due to time constraints, I think I am going to have to kick the can down the road and let [FS 1048](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1048-ce-builder-map.md) handle this. I have sketched most of it out on top of the changes in this RFC and it looks fine to do (although I haven't given `use! ... return ...` all the attention it deserves).

PR #341, opened 2018-10-24

dsyme:
I've made a bunch of edits to the RFC for nullable reference types to try to make it more along the lines of a specification of the proposed F# language rules for the feature.

Roughly speaking this means additions that would be suitable to the style of the F# language specification, focusing on the F# language rules and definitions.


dsyme:
@cartermp 
* I've eliminated quite a lot of the discursive text aince we can just dive in and define the constructs and rules. 

* I've added some library helpers that are present in the prototype in preliminary form.  These aren't at all final but illustrate some of the issues with the feature

* I've moved quite a lot of the "possibly" sections to the "alternatives" section since I don't think we'll actually be doing them. We can reconsider as time goes on but I want to iterate to something that roughly corresponds to the prototype

PR #343, opened 2018-10-31
Create FS-1066  Add tryExactlyOne to array, list and seq

gdziadkiewicz:


RFCs/FS-1066-tryExactlyOne.md
```diff
@@ -0,0 +1,91 @@
...
+Seq.tryExactlyOne (Seq.ofList [1; 2])
+// None
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Extending api surface.
+
+# Alternatives
```
cartermp: Another alternative to list here would be to return `None` when `null` is passed as an argument. If you could list the tradeoff of doing that it would be good.

RFCs/FS-1066-tryExactlyOne.md
```diff
@@ -0,0 +1,91 @@
...
+Seq.tryExactlyOne (Seq.ofList [1])
+// Some 1
+Seq.tryExactlyOne (Seq.ofList [1; 2])
+// None
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Extending api surface.
```
cartermp: I'd also add that because this has different behavior from the LINQ equivalent, C# developers learning expecting something to throw may be a bit confused.

(Though I think most people will agree that returning `None` is more idiomatic)

RFCs/FS-1066-tryExactlyOne.md
gdziadkiewicz: Added it.

RFCs/FS-1066-tryExactlyOne.md
gdziadkiewicz: Added it.

PR #350, opened 2019-03-12
Create FS-1068-open-static-classes.md

dsyme:


duncanmak:
I wonder if we should consider this along with this FS: https://github.com/fsharp/fslang-suggestions/issues/657.

I have a patch to the compiler here: https://github.com/Microsoft/visualfsharp/pull/4673.

PR #359, opened 2019-04-16
Add FST-1008

cartermp:
Here's the proposal to migrate FSharp.Compiler.Service to `dotnet/fsharp` and archive the current repository.

- [x] Add section on compatibility rules (hint: it'll be semantic versioning with no binary compat)
- [x] Consider Fable fork
- [x] Goals for development of FCS without Windows or VS dependencies

### DIFF
--- a/tooling/FST-1008-publish-fcs-from-upstream-repo.md
+++ b/tooling/FST-1008-publish-fcs-from-upstream-repo.md
@@ -56,3 +56,12 @@ The README must be modified to state clearly that this repository is archived, a
 Additionally, we should retain the historical context of this repository to ensure that we don't lose any of the rich history it has developed over the years.
 
 Finally, once all of the above is completed, we can archive the repository.
+
+## Compatibility and versioning
+
+Despite the package being at version `28.0.0` at the time of writing, it's best to think of it as version `0.28`. It is not binary compatible and is subject to many breaking changes. The FCS API itself is likely to undergo more changes in the future, and consumers of that API will have to adjust to those changes. This is the norm today, and we expect it to continue for quite some time. Given this, the compatibility and versioning rules are:
+
+* The binary in the package is **not** binary compatible and will change
+* The package will follow semantic versioning, where any major version change indicates a possible breaking change
+
+In the future, the FCS APIs may stabilize and binary compatibility may be enforced. But that is at an unknown time long in the future.
--- a/tooling/FST-1008-publish-fcs-from-upstream-repo.md
+++ b/tooling/FST-1008-publish-fcs-from-upstream-repo.md
@@ -1,0 +1,58 @@
+# FST 1008 Publish FCS from the dotnet/fsharp repository
+
+This RFC proposes that we publish the [FSharp.Compiler.Service](https://www.nuget.org/packages/FSharp.Compiler.Service/) package from the main F# development repository (`dotnet/fsharp`) and archive the [current downstream repository](https://github.com/fsharp/fsharp.compiler.service).
+
+## Publishing the package from dotnet/fsharp
+
+### Signing
+
+For starters, as per Microsoft policy, any official package considered to be a part of the F# product must be signed:
+
+* The binaries must be signed by Microsoft signing keys
+* The package itself must be signed by Microsoft signing keys
+
+This would be the same as FSharp.Core today.
+
+### Prerelease usage
+
+The F# build would produce a signed FSharp.Compiler.Private package. This would be accessible in a build artifacts directory for anyone looking to build this package against a particular branch of the F# codebase.
+
+Additionally, Microsoft will publish a nightly feed on MyGet where a new package would be published every time there is a signed F# build.
+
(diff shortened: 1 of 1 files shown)


cartermp:
Soliciting feedback from:

* @enricosada 
* @Krzysztof-Cieslak
* @baronfel 
* @auduchinok
* @ninjarobot

0x53A:
What do you think about adding a version postfix to the binary, so that multiple versions of F.C.S can be loaded into one process without conflicts?

I actually haven't seen this pattern yet in .net, but it is common in the native world.

So the next version would be ``FSharp.Compiler.Service.29.dll``.

Krzysztof-Cieslak:
Before getting to any feedback I'll add @ncave and @alfonsogarciacaro to this discussion as it impacts Fable fork.

ncave:
IMO the impact on Fable would be minimal.
Can't comment on other aspects of moving the repo, I'll defer to the maintainers of FCS.

cartermp:
@ncave I believe moving to a `dotnet/fsharp` repo would require a fork of the entire repo, which is more stuff to keep on top of. We could set up a bot that attemps to auto-merge whenever there are no conflicts, but the total code churn will be higher. So that could be a downside.

ncave:
@cartermp
Yes, you are right that the churn will be higher, but at least it provides access to the latest F# features. Technically to control churn we can switch Fable support from a PR on `dotnet/fsharp` (which will probably never get merged anyway) to a PR on a fork of `dotnet/fsharp` that gets updated only on demand.

I don't see a problem with that, in fact I personally like the move for reasons stated above, as long as cross-platform FCS development and tests don't require installing too many dependencies (it's a bit hard to contribute to the compiler right now because of all the prerequisites).

cartermp:
@ncave Thanks! I've added two sections - let me know if you have any objections.

ncave:
@cartermp Looks good, perhaps add "Thou shalt be able to run the FCS tests" to the build section, unless it's assumed. (Sorry, couldn't resist the pun, it was just hanging there :).

tooling/FST-1008-publish-fcs-from-upstream-repo.md
```diff
@@ -89,5 +89,6 @@ To reiterate - to build FCS in `dotnet/fsharp`:
 
 * You should only require an OS that .NET Core can run on; i.e., no Windows-only dependencies
 * You should not need to install any machine dependencies to build FCS
+* You should be able to run all FCS tests without machine dpeendencies or a particular OS
```
ncave: @cartermp Small typo there.

enricosada:
Looks good to me, awesome to move forward to a single repo :+1:

@cartermp i'd like to add some notes on dev flow, to make shared maintenance easier also for outside MS, and maintain current features.

Each point is independent.

1. maintain git tag for all stable releases, like [now](https://github.com/fsharp/FSharp.Compiler.Service/releases)
2. maintain the use of a [RELEASE_NOTES.md](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/fcs/RELEASE_NOTES.md) like now, because is useful for consumers
3. no nightly, just signed prereleases from `master` branch to a myget feed
4. maintain a shared ownership to publish a stable, using git tag on `master` who generate a signed package.

### git tags for stable

now all stable release are tagged, and that help a lot for regression etc.
we cannot use same naming convention (`27.0.1`) but something like `FCS-27.0.1` or `FCS-v27.0.1` can be ok so doesnt collide with others tags.

### use a RELEASE_NOTES.md

the [RELEASE_NOTES.md](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/fcs/RELEASE_NOTES.md) is really nice and help track changes at hight level, instead of read all the commits
Hopefully the maintainer should update that as needed, but every stable release should have it.
As a note, it's now used to generate the version number itself.

### prerelease flow

For me it's ok that just Microsoft can build the signed package, it mean it was built by MS CI and signed.

I think, if possible, we dont need the F# extension flow with nightly (that's strange for a package).

I think we just need a prerelease package for each commit in `master` branch, so is automatic for every green build, and zero maintenance.

If a commit is merged in master, mean has passed the approval of maintainers.

The signed package (as prerelease) can be built by the MS owned CI every commit in master and pushed to a myget feed.

### shared ownership for publish a stable

If possible, having two different way of publishing (unsigned and signed), mean packages are different.

I think is possible to allow to maintain the publishing from community and the requirement for signing like:

- a build with a git tag like `FCS-v27.0.0` trigger a CI job on MS
- in that build job, the package is generated without prerelease AND signed
  - optionally the build can be set as failure if RELEASE_NOTES.md is not up to date with release date.
- the package after that pushed on nuget.org, automatically or with manual approval

Using Azure Devops is possible to do all that, having a different job for tags, and a manual `Release` job to publish (using secrets in azure devops)

Like that you can add two release jobs with different auth (MS and FSF), and each one use a different nuget api key, so is known the publisher (MS or FSF) but the package is signed

That mean community (the person choosed by FSF, not everyone) can trigger a new signed build of the package, using MS infrastructure for signing.
that mean the package is really co-owned i think

PR #372, opened 2019-07-05
Add DIMs first draft

cartermp:
Needs:

- [x] Go over samples again to make sure they're accurate

### DIFF
--- a/RFCs/FS-1074-default-interface-member-inteop.md
+++ b/RFCs/FS-1074-default-interface-member-inteop.md
@@ -96,10 +96,6 @@ let i' =
 printfn "%d" i'.GetNumber() // 13
 ```
 
-## Reabstracting through inheritance
-
-TODO - only applicable when inheriting from an interface with a DIM on it.
-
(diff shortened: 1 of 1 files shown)


dsyme:
> There's a long comment I wrote up somewhere on the original design thread for C# which we should reference here. I'll try to find it

Ah it's referenced here in the language suggestion: https://github.com/fsharp/fslang-suggestions/issues/679#issuecomment-399411192

RFCs/FS-1074-default-interface-member-inteop.md
```diff
@@ -0,0 +1,138 @@
...
+## Overriding a default implementation
+
+It is possible to override a default implementation when implementing an interface in F# code:
+
+```fsharp
+open Dims
+
+type C() =
+    interface IHaveADefaultMember with
+        override __.GetNumber() = 13
```
TIHan: `override` is not what we want. Using `member` is fine. C# has refrained from using `override` word.

PR #379, opened 2019-08-15
Update open static classes RFC

cartermp:
- [x] Convert to a full design doc
- [x] Add method overloading
- [x] Add rules for resolving methods vs. properties
- [x] Add section on respecting attributes
- [x] Open question for opening generic static classes

Remaining:
- [x] More examples
- [x] Double-check name resolution rules
- [x] Mention type providers

### DIFF
--- a/preview/FS-1068-open-static-classes.md
+++ b/preview/FS-1068-open-static-classes.md
@@ -186,7 +186,7 @@ M(1)
 
 ### Background: name resolution in the context of a type
 
-Type-directed member resolution in F# prefers properties over methods. Section 14.1.5 states:
+Type-directed member in section 14.1.5 states:
 
 > Name Resolution for Members is a sub-procedure used to resolve `.member-ident[.rest]` to a member, in the context of a particular type `type`.
 > ...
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
@cartermp this looks great, I have a question on section

> Only static classes can be opened
...
> This functionality is explicitly scoped out for now.

This seems contradictory with an important aspect of the feature, or maybe I am misunderstanding that part?

The remaining of the RFC shows usage of open, maybe rephrasing this way:

This restriction is not part of the RFC, using `open` is valid on classes whether or not they are defined as static.

cartermp:
> This seems contradictory with an important aspect of the feature, or maybe I am misunderstanding that part?

This was explicitly scoped out early on, the old design notes make mention of it as well.

dsyme:
>   Add method overloading

Not allowing combination of unrelated method overload sets was a deliberate design decision, but we can discuss further (I notice you've updated the RFC with your intended change).

Personally, from a design perspective I prefer not combining unrelated overload sets - there is a lot of hell and misuse awaiting future users of libraries that make use of this feature.  

However I can see why you've changed it, and from an API interop perspective it makes sense to align with C#.

cartermp:
> Personally, from a design perspective I prefer not combining unrelated overload sets - there is a lot of hell and misuse awaiting future users of libraries that make use of this feature.

This we'll have to discuss. It also has ramifications for generic static classes and consumption of APIs that use the as well. Simply having the last one shadow everything else feels very odd.

Additionally, doing this would be similar to bringing type extensions into scope. Though it's not 100% the same mechanism, it "feels" appropriate since we have an analogue in the language today.

dsyme:
@cartermp I'm actually OK with the proposed resolution of this issue by yourself and @TIHan.

PR #384, opened 2019-08-16
Update FS-1068-open-static-classes.md

dsyme:
This removes some discussion which was misleading.  

* The `open static C<int>` feature genuinely adds expressivity that is potentially relevant to F#.

* The discussion about some F# code not being possible to write in F# was misleading

### DIFF
--- a/preview/FS-1068-open-static-classes.md
+++ b/preview/FS-1068-open-static-classes.md
@@ -279,28 +279,6 @@ M(12); // This is one overload
 M("hello"); // This is another overload
 ```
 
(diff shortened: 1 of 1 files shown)


cartermp:
Note that we'll also have to specify:

* Whether or not `open C<_>` is allowed
* How to resolve `M(x: int)` and `M(x: 'T)` where `'T` is is `int` from an `open C<int>` declaration
* If there are other analogues here in other languages/environments and if that impacts expressivity

dsyme:
@cartermp Please add those to the section in the RFC.  I think however we should not try to implement `open C<int>` as part of this RFC, it just adds too much complication and we don't yet have a use case of an API which is forcing us to do this.

PR #387, opened 2019-08-16
Add named entity to alternatives for nameof

cartermp:


cartermp:
Also clean up a few thangs

PR #389, opened 2019-08-21
Adjust RFC for "open type" and other resolution of design issues

dsyme:
Switch to `open type` and note where RFC doesn't match

Move resolved issue to "alternatives"

### DIFF
--- a/preview/FS-1068-open-static-classes.md
+++ b/preview/FS-1068-open-static-classes.md
@@ -14,20 +14,18 @@ This RFC covers the detailed proposal for this suggestion.
 ## Summary
 [summary]: #summary
 
-Add support for opening static classes, e.g. `open System.Math`. For example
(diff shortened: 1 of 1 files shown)


preview/FS-1068-open-static-classes.md
```diff
@@ -176,35 +182,59 @@ N(12.0) // Works!
...
+open type System.Numerics.Vector2
 
 let v1 = Vector2(1.0f, 2.0f)
 let v2 = Vector2(1.0f, 2.0f)
 
 Dot(v1, v2) // No need to fully qualify 'Dot'
 ```
 
-But it is currently considered out of scope.
+In the preview only static classes could be opened. Post preview have resolved to allow non-static classes to be opened.
```
cartermp:
```suggestion
In the preview only static classes could be opened. Post preview have resolved to allow non-static types to be opened.
```

Note that we should also support opening structs like `Vector2`

PR #392, opened 2019-09-18
Slicing for 3D/4D arrays with fixed index

nelson-wu:
This is to enable syntax like `arr3d.[0, 0, 0..1]`.


dsyme:
Great, thank you!

nelson-wu:
@dsyme You're very welcome! 😁

PR #400, opened 2019-09-24
From the end slicing for collections

nelson-wu:
This RFC proposes the addition of the `^` symbol in the current slicing syntax to index slices from the end of a collection.

### DIFF
--- a/RFCs/FS-1076-from-the-end-slicing.md
+++ b/RFCs/FS-1076-from-the-end-slicing.md
@@ -1,0 +1,86 @@
+# F# RFC FS-1076 - From the end slicing for collections
+
+The design suggestion [FILL ME IN](https://github.com/fsharp/fslang-suggestions/issues/fill-me-in) has been marked "approved in principle".
+This RFC covers the detailed proposal for this suggestion.
+
+* [x] Approved in principle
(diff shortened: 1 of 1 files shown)


cartermp:
This will also have to make mention of how we handle `System.Index` and `System.Range` given that they're only in .NET Core 3.0 and we cannot take a simple dependency on them in FSharp.Core.

nelson-wu:
Right now this RFC is slightly outdated in that it operates on the existing API signature for GetSlice, which through the 3D slicing PR we've found to be somewhat inadequate for our needs. If we choose to use Range/Index, we're definitely not using the existing API for GetSlice, which means the implementation of from-the-end slicing is going to be somewhat different from what is contained here.

nelson-wu:
@dsyme @KevinRansom @cartermp Me and @TIHan came up with an updated design that removes the dependency on Length. Could you please look it over again?

dsyme:
> @dsyme @KevinRansom @cartermp Me and @TIHan came up with an updated design that removes the dependency on Length. Could you please look it over again?

I like it!

dsyme:
I will merge this, let's iterate from there

PR #406, opened 2019-10-23
Update and-bang RFC

dsyme:
Update with simplified design for `and!`

### DIFF
--- a/RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
+++ b/RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
@@ -1,4 +1,4 @@
-# F# RFC FS-1063 - Support let! .. and... for simultaneous bind to reduce re-execution in computation expressions
+# F# RFC FS-1063 - Support and! for simultaneous bind in computation expressions
 
 The design suggestion [Support let! .. and... for applicative functors](https://github.com/fsharp/fslang-suggestions/issues/579) has been marked "approved in principle".
 This RFC covers the detailed proposal for this suggestion.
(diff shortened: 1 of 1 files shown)


RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
```diff
@@ -242,730 +208,30 @@ ce {
...
 # Drawbacks
 [drawbacks]: #drawbacks
 
-The new applicative computation expressions are quite constrained, and as has been discussed, that is precisely what allows them to be so useful. However, these constraints are potentially somewhat unintuitive to the beginner. Computation expressions already involve one of the steeper learning curves of the F# language features, so the added complexity from this feature needs to be carefully weighed against the potential guarantees, expressiveness and performance gains that they can offer.
+TBD
 
 # Alternative Designs
 [alternative-designs]: #alternative-designs
 
+We chose not to support `use!` or `anduse!`. Why: TBD
```
cartermp: The `TBD` here will block on merging the feature, since we'll need a stated reason

cartermp:
Why are we removing examples that don't involve `use!` or `anduse!`?

RFCs/FS-1063-support-letbang-andbang-for-applicative-functors.md
dsyme: Yes I know

dsyme:
> Why are we removing examples that don't involve use! or anduse!?

Most of them are illustrating the previous syntactic restrictions.  The ones that aren't correspond to other simple examples.  Also the RFC doesn't have to be a full exposition of applicatives now the feature is much simpler.

PR #409, opened 2019-10-24
Add BindReturn and example of dependency graphs

dsyme:


dsyme:
This incorprates https://github.com/fsharp/fslang-suggestions/issues/36 into the RFC, which gives satisfying overall results for dependency graphs, giving zero node re-allocation on recomputation for the static portion of graphs, while still allowing dynamic parts of graphs.  This is really the core result we are after (for various variations on dependency graphs)

PR #412, opened 2019-11-05
Update from the end slicing RFC to include indexing

nelson-wu:
Decided while implementing that it wouldn't make sense if we supported this only in slice and not in index.

### DIFF
--- a/RFCs/FS-1076-from-the-end-slicing.md
+++ b/RFCs/FS-1076-from-the-end-slicing.md
@@ -123,7 +123,7 @@ This would break any code that defines `..^` as a custom operator.
 
 ## Third-party collections
 
-We currently require third party collections to implement `<'T>.GetSlice` to support slicing syntax and `<'T>.Item` for indexing. We can additionally require third party collections to implement `<'T>.GetReverseIndex` if they wish to support from-the-end indexing and slicing. If they choose to do only the former, any `^i` indices will fail with a compile time error.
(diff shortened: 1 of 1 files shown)


RFCs/FS-1076-from-the-end-slicing.md
```diff
@@ -123,7 +123,7 @@ This would break any code that defines `..^` as a custom operator.
 
 ## Third-party collections
 
-We currently require third party collections to implement `<'T>.GetSlice` to support slicing syntax and `<'T>.Item` for indexing. We can additionally require third party collections to implement `<'T>.GetReverseIndex` if they wish to support from-the-end indexing and slicing. If they choose to do only the former, any `^i` indices will fail with a compile time error.
+We currently require third party collections to implement `<'T>.GetSlice` to support slicing syntax and `<'T>.Item` for indexing. We can additionally require third party collections to implement `<'T>.GetReverseIndex` if they wish to support from-the-end indexing and slicing. If they choose to do implement only `GetSlice`/`Item`, any `^i` indices will fail with a compile time error.
```
abelbraaksma: My point was mainly with the confusion coming from 'we can additionally...' (meaning, this is an option and there's no agreement yet), and then the next sentence 'if they choose...' (which suggests that `GetReverseIndex` must be implemented).

The choice made now is important as we cannot change it later. I think it's best to choose either:

* `GetReverseIndex` is always a requirement of this syntax is used
* or `GetReverseIndex` is used if it's available, otherwise `GetSlice/Item` will be used
* or we scrap `GetReverseIndex` for now, and allow such optimisation later, in which case it can never become a requirement, but at most a compiler preference

(I should admit that I didn't take a deep look at the state of implementation, it's possible that a choice is already made, in which case that choice should be reflected here) 

Also: s/to do implement only/to only implement/ ;)

abelbraaksma: I just reread the whole text and it seems that it was already decided that `GetReverseIndex` is required on 3rd party types. It's just that only in this section it's written as if it's an optional design choice that still needs to be made.

PR #420, opened 2020-01-06
Create FS-1080-extend-fixed-expressions

jwosty:
grab the RFC number - will fill in more later


RFCs/FS-1081-extend-fixed-expressions.md
```diff
@@ -0,0 +1,55 @@
...
+* [x] Approved in principle
+* [ ] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/fill-me-in)
+* [ ] Details: [under discussion](https://github.com/fsharp/fslang-design/issues/FILL-ME-IN)
+* [ ] Implementation: [In progress](https://github.com/Microsoft/visualfsharp/pull/FILL-ME-IN)
+
+
+# Summary
+[summary]: #summary
+
+The `fixed` expression should be extended `ref` types and any type that implements the C#-style [`GetPinnableReference()`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.3/pattern-based-fixed) pattern.
```
cartermp:
```suggestion
The `fixed` expression should be extended to support `byref` types and any type that implements the C#-style [`GetPinnableReference()`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.3/pattern-based-fixed) pattern.
```

RFCs/FS-1081-extend-fixed-expressions.md
```diff
@@ -0,0 +1,55 @@
+# F# RFC FS-1081 - Extend fixed expressions to support ref types and types implementing GetPinnableReference()
```
cartermp:
```suggestion
# F# RFC FS-1081 - Extend fixed expressions to support byref types and types implementing GetPinnableReference()
```

jwosty:
@cartermp's reviews of changing "ref" to "byref" brings up an interesting point: I wonder if fixed should also be able to be used with `ref<'a>`. If we kept it just to `byref<'a>` (and GetPinnableReference), which is what the current draft perscribes, I think you could still pin a ref's contents like so:
```let x = ref 42
use ptr = fixed &x```

Could you not? Why not make this easier, since F# seems to encourage refs over mutable byrefs when possible?

**Comment by cartermp** [comment]
@jwosty it's actually the inverse, we want to discourage the use of `ref` unless it's absolutely necessary.

File: RFCs/FS-1081-extend-fixed-expressions.md [hunk 3]
```diff
@@ -0,0 +1,55 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This feature would introduce more special rules into the language.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Don't do this, providing no alternative for those who want to use Span<'T> and friends with native code from F#
+- Only extend `fixed` to support `ref` types
```
cartermp: What does this mean? Since `Ref<'T>` doesn't support `GetPinnableReference` it doesn't seem related to this work. Reference cells aren't mentioned elsewhere in this doc.

RFCs/FS-1081-extend-fixed-expressions.md
```diff
@@ -0,0 +1,55 @@
...
+- Only extend `fixed` to support `GetPinnableReference()`
+  
+# Compatibility
+[compatibility]: #compatibility
+
+Please address all necessary compatibility questions:
+* Is this a breaking change?
+* What happens when previous versions of the F# compiler encounter this design addition as source code?
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+* If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct?
```
cartermp: These need to be answered.

RFCs/FS-1081-extend-fixed-expressions.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+It should be possible to pin and take native references of types like `Span<'T>` and `Memory<'T>` via `fixed` statements, as this is useful especially in native interop scenarios. At the moment, there are no workarounds in pure F#: you can do this in C#, but it's impossible in F#. The best you can do is write that part of your code in C#.
+
+# Detailed design
+[design]: #detailed-design
+
+This section of the RFC needs more detail and specific examples. Please feel free to contribute by editing this file and submitting a PR to fsharp/fslang-design.
```
cartermp: This needs to be deleted.

RFCs/FS-1081-extend-fixed-expressions.md
```diff
@@ -0,0 +1,55 @@
...
+
+This will require both type-checker and codegen changes. Statements in the following form: `use ptr = fixed expr` now need to successfully type-check for any `expr` where its type is a `byref<'a>` type or something that implements `GetPinnableReference()`. The code generator will need to generate code that pins these references and takes their addresses. For reference, here is the C# proposal: https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.3/pattern-based-fixed.md.
```
cartermp: It would be good to explain that `fixed` can only be used when `expr` is an array, string, address of an array element, or an address of a field today. This would add support for any `byref` or something with the `GetPinnableReference()` member defined. Formatting that as a bullet-pointed list would help with readability.

RFCs/FS-1081-extend-fixed-expressions.md
jwosty: Good catch, I missed that one when changing all the refs to byrefs

dsyme:
Great work on this :-)

PR #437, opened 2020-02-13
Update the custom attribute code in FS-1053-span.md.

teo-tsirpanis:
Some notes about the changes:

* It is better for the attributes to become internal, as they will pollute the public API of a library. This is what C# does as well.
* `IsByRefLikeAttribute` is defined to be only allowed to be placed on structs.
* `AttributeUsage.AllowMultiple` is false by default.
* Some unused namespaces and fully qualified names were removed.

### DIFF
--- a/FSharp-4.5/FS-1053-span.md
+++ b/FSharp-4.5/FS-1053-span.md
@@ -570,20 +570,21 @@ type DateTime with
 ```
 This makes it impossible to define byref extension properties in particular. 
 
-* Note that `IsByRefLikeAttribute` is only available in .NET 4.7.2.
+* Note that `IsByRefLikeAttribute` is only available in at least .NET 4.7.2 or .NET Standard 2.1.
+
+Those who are targeting other frameworks can define these attributes in their own code this way:
 
 ```fsharp
 namespace System.Runtime.CompilerServices
     open System
-    open System.Runtime.CompilerServices
-    open System.Runtime.InteropServices
-    [<AttributeUsage(AttributeTargets.All,AllowMultiple=false); Sealed>]
-    type IsReadOnlyAttribute() =
-        inherit System.Attribute()
 
-    [<AttributeUsage(AttributeTargets.All,AllowMultiple=false); Sealed>]
-    type IsByRefLikeAttribute() =
-        inherit System.Attribute()
+    [<AttributeUsage(AttributeTargets.All); Sealed>]
+    type internal IsReadOnlyAttribute() =
+        inherit Attribute()
+
+    [<AttributeUsage(AttributeTargets.Struct); Sealed>]
+    type internal IsByRefLikeAttribute() =
+        inherit Attribute()
 ```
 
 


cartermp:
I actually think that this section should be removed entirely, since this fundamental runtime concept is not understood by .NET Framework (and never will be) we shouldn't even call it out. @dsyme?

teo-tsirpanis:
@cartermp, it might not be understood from .NET Framework, but is understood from the compilers who enforce some restrictions, even on .NET Framework.

And as I said in dotnet/fsharp#8564, it is useful for those who write .NET Standard 2.0 libraries.

cartermp:
I disagree. There's no benefit to sharing this kind of code.

teo-tsirpanis:
> [...] sharing this kind of code.

You mean ref structs in a public API? Besides spans, for example [`System.Text.Json.Utf8JsonReader`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader?view=netcore-3.1) is a ref struct as well.

I agree that ref structs are for low-level, performance-critical scenarios (essentially for types holding spans) and have no place in the public API of higher-level libraries.

cartermp:
System.Text.Json is only for use with .NET Core or Xamarin. The scenario mentioned in this RFC is about allowing the use of byref-like structs in a .NET Framework environment. This is why I don't think it's worth mentioning in the RFC at all.

teo-tsirpanis:
> The scenario mentioned in this RFC is about allowing the use of byref-like structs in a .NET Framework environment.

And in .NET Standard 2.0. 🙂

teo-tsirpanis:
Or are you suggesting to just remove _the mention_ of .NET Framework in the document and keep it for .NET Standard?

That would seem more reasonable, but I would still disagree, because in C# you can create ref structs from any version of the .NET Framework (tested with .NET Framework 2.0), and the compiler does not allow you to place them on the heap, even though you can still do that by getting your hands dirty with IL.

cartermp:
Sorry for the delay here. I think it should be removed entirely, including when considering .NET Standard 2.0.

Byref-like structs, like Span, are fundamentally a "modern .NET" concept. They are not implemented nor understood by the .NET Framework (desktop .NET), and that fact will never change because the desktop .NET runtime is feature complete.

The purpose of .NET Standard 2.0 is to share code between .NET Core and .NET Framework. So when you have a construct that .NET Framework does not understand, that scenario falls apart.

The relevant attributes were introduced for code sharing in .NET Standard 2.1, which is explicitly incompatible with the .NET Framework. When sharing code between Xamarin and .NET Core today, you target .NET Standard 2.1 and get these attributes anyways. There's no need to define them.

This is why I don't think it's worth even mentioning this in the first place.

dsyme:
> I actually think that this section should be removed entirely, since this fundamental runtime concept is not understood by .NET Framework (and never will be) we shouldn't even call it out. @dsyme?

agreed

dsyme:
I'll close this per discussion above

PR #442, opened 2020-03-07
Create FS-1084-The-UnknownEnum-pattern.md

Happypig375:
An official approved-in-principle still hasn't yet been received, but I believe that this addition meets the bar to be approved.

### DIFF
--- a/RFCs/FS-1084-The-UnknownEnum-pattern.md
+++ b/RFCs/FS-1084-The-UnknownEnum-pattern.md
@@ -23,20 +23,13 @@ An `UnknownEnum` pattern would be an ideal solution to this, as it matches all c
 
 # Detailed design
 
-A new active pattern, named `UnknownEnum` will be added to FSharp.Core:
-```fs
-type private UnknownEnumLookup<'Enum>() =
-    static member val Values = typeof<'Enum>.GetEnumValues() :?> 'Enum[]
-let (|UnknownEnum|_|) (enum:'Enum when 'Enum : enum<'Underlying>) =
-    if Array.IndexOf(UnknownEnumLookup<'Enum>.Values, enum) < 0 then
-        Some <| LanguagePrimitives.EnumToValue enum
-    else None
-```
-This pattern deconstructs the underlying value of the enum for concise logging and error reporting.
+A new active pattern, named `UnknownEnum` will be added to FSharp.Core.
 
-The compiler will need to be updated to recognize this pattern and consider all unknown enum cases handled in match completeness analysis. As a result, when this pattern is used, FS0104 will be silenced.
+In the compiler's view, it will match all enum members statically available. Therefore, the compiler will need to be updated to recognize this pattern and consider all unknown enum cases handled in match completeness analysis and IL generation. As a result, when this pattern is used, FS0104 will be silenced.
 
-`match` expressions on enums can use this pattern:
+This pattern will also deconstruct the underlying value of the enum for concise logging and error reporting.
+
+`match` expressions on enums can then use this pattern:
 ```fs
 type E = A = 1 | B = 2
 // No warnings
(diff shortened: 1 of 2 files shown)


abelbraaksma:
Good to have the proposal, though the discussion wasn't fully completed yet, with still some open questions and alternative solutions. But perhaps this helps focusing on the actual proposal ;).

I've skimmed the text, but have yet one small request: could you write the code examples in F# style according to the guidelines (mainly, each `|` on a newline, aligned with the m of match or the t of type)? They're a bit hard to read and understand this way.

I think we should drop exhaustive testing when `Flags` are used. Typically, such enums are or'ed or and'ed to find out set flags and rarely use the individual cases purely.

Happypig375:
@abelbraaksma So do you support warning or erroring when flag enums are encountered?

abelbraaksma:
I think the predictable thing to do is to give a standard warning that there are more matches. We don't have to fully disallow its use for flags. I don't know what the core functions return when you enumerate all enum fields.

There's another important corner car, actually two:

* multiple enum fields can have the same value
* with flags, fields cab overlap, or be defined as a mask (you often see one field `0xFFFFFFFF`,to make it easy to cover all flags, and to still make the enum expandable).

We need some way to wield out those duplicates,which should not be considered distinct fields.

Happypig375:
>I think the predictable thing to do is to give a standard warning that there are more matches. We don't have to fully disallow its use for flags.

So you support the alternative proposal?
>I don't know what the core functions return when you enumerate all enum fields, though, you don't want to accidentally give false positives (that is, the user has to be confident that no warning means all flags are covered).

It's all the defined values, consistent with common sense.
```fs
> type [<System.Flags>] E = A = 1 | B = 2 | C = 4 | D = 8;;  
type E =
  | A = 1
  | B = 2
  | C = 4
  | D = 8

> typeof<E>.GetEnumValues();;
val it : System.Array =
  [|A {value__ = 1;}; B {value__ = 2;}; C {value__ = 4;}; D {value__ = 8;}|]

>  it :?> E[] |> Array.reduce(|||);;
val it : E = A, B, C, D {value__ = 15;}

> ~~~it;;
val it : E = -16 {value__ = -16;}
```
>We could consider an additional warning when used with flags, I'm not sure there.

So you don't actually support the alternative proposal?
>multiple enum fields can have the same value

That's handled fine in this implementation. The match completeness checking will need to watch out for this and issue a duplicate match warning, though.
>with flags, fields can overlap, or be defined as a mask (you often see one field 0xFFFFFFFF,to make it easy to cover all flags, and to still make the enum expandable).

That makes `UnknownEnum` on flag enums ambiguous. That is an additional reason not to implement the alternative proposal.
>what happens when different enums are mixed. And what if the input to the match is an int (maybe the type system catches that, didn't check).

```fs
[<Flags>]
type E = A = 1 | B = 2 | C = 4 | D = 8
type F = Incompatible = 12
match enum<E> 3 with
// FS0001: This expression was expected to have type 'E' but here has type 'F'
| HasFlag F.Incompatible -> 0
| _ -> 1
```
```fs
[<Flags>]
type E = A = 1 | B = 2 | C = 4 | D = 8
type F = Incompatible = 12
match 3 with
// FS0001: This expression was expected to have type 'int' but here has type 'F'
// FS0001: The type 'int' is not compatible with the type 'Enum'
| HasFlag F.Incompatible -> 0
// FS0001: The type 'int' is not a CLI enum type
| UnknownEnum x -> x
| _ -> 1
```

Happypig375:
`HasFlag` checks if the value has all the flags provided.
```fs
> E.A.HasFlag (E.A ||| E.B);;
val it : bool = false

> E.B.HasFlag (E.A ||| E.B);;
val it : bool = false

> (E.A ||| E.B).HasFlag (E.A ||| E.B);;
val it : bool = true
```

Happypig375:
@dsyme @cartermp Still no sign of merging yet?

cartermp:
In this case I'd want @dsyme to have a say re: approval on the suggestion before merging here.

RFCs/FS-1084-The-UnknownEnum-pattern.md
```diff
@@ -0,0 +1,127 @@
...
+An `UnknownEnum` pattern would be an ideal solution to this, as it matches all cases that FS0104 would warn against, while keeping FS0025 warnings when a new case is added to the enum type. As a result, using `match` expressions on enums can be both safe, as in matching all potential values, and future-proof, as in keeping FS0025 warnings when a new case is added.
+
+# Detailed design
+
+A new active pattern, named `UnknownEnum` will be added to FSharp.Core:
+```fs
+type private UnknownEnumLookup<'Enum>() =
+    static member val Values = typeof<'Enum>.GetEnumValues() :?> 'Enum[]
+let (|UnknownEnum|_|) (enum:'Enum when 'Enum : enum<'Underlying>) =
+    if Array.IndexOf(UnknownEnumLookup<'Enum>.Values, enum) < 0 then
```
dsyme, line 31: This implementation is based on the **runtime** set of enum values.  Shouldn't it be based on the static set?

Happypig375: Updated the document

PR #452, opened 2020-03-25
Renumber nameof pattern RFC to 1085 + add discussion link

Tarmil:
Renumber due to clash with [RFC-1084](https://github.com/fsharp/fslang-design/pull/442).


cartermp:
RFC-1084 isn't approved yet, though, so I don't think that will be merged in unless that happens.

dsyme:
> RFC-1084 isn't approved yet, though, so I don't think that will be merged in unless that happens.

I'm ok with there being gaps in the sequence!  Compactness is for nerds :)

PR #453, opened 2020-04-01
Create FS-1087-resumable-code-and-task-builder

dsyme:
First notes on resumable code and `task { ... }` RFC


Happypig375:
Now we magically skip 1086?

dsyme:
> Now we magically skip 1086?

"Knut IV, the Saint, king of Denmark (1080-86), murdered" :-)

In truth, I couldn't work out what the right number was from 1084/1085 renumbering thing, so just skipped up one

PR #460, opened 2020-04-15
RFC FS-0001 - drop the use of string interpolation with FormattableString or IFormattable?

dsyme:
Updates proposing to drop the use of string interpolation with FormattableString or IFormattable and push that off to a separate RFC (since it also affects the use of existing printf format strings with these two types)

See comment at https://github.com/fsharp/fslang-design/issues/368#issuecomment-614160241 for discussion. I won't merge right away.

Plus some other cleanup.


dsyme:
I will add back in support for FormattableString separately

PR #462, opened 2020-05-01
Update FS-1001-StringInterpolation.md

dsyme:
I added details on using format strings with `printf` to the RFC by direct commit

Also adding this clarification

### DIFF
--- a/RFCs/FS-1001-StringInterpolation.md
+++ b/RFCs/FS-1001-StringInterpolation.md
@@ -102,6 +102,17 @@ Use cases include any for which printf and variants are currently used: console 
 A mix of type-checked and unchecked fills **is** allowed in a single format string when typed as type `string`. For example `$" abc %d{3} def {5}"` is allowed.
 
 
+### Activation
+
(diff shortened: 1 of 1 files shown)


RFCs/FS-1001-StringInterpolation.md
```diff
@@ -102,6 +102,17 @@ Use cases include any for which printf and variants are currently used: console
 A mix of type-checked and unchecked fills **is** allowed in a single format string when typed as type `string`. For example `$" abc %d{3} def {5}"` is allowed.
 
 
+### Activation
+
+The feature is only activated when an FSharp.Core library supporting the feature is referenced at compile-time.  This is determined
```
cartermp, line 107: Hmmm. That doesn't seem right. What about purely .NET-style interpolated strings? This would require a feature flag in the compiler as well, activated by `preview` or the final explicit or implicit LangVersion when this is shipped

dsyme: Sorry, yes, it's also gated on compiler `--langversion:preview`

PR #466, opened 2020-05-18
FS-1088 - if! (if-bang)

jwosty:


dsyme:
@jwosty We should add `while!` at the same time I guess.

jwosty:
@dsyme Alright, I will be adding that in.

smoothdeveloper:
@dsyme I think it is good to merge "add RFC file" PRs (helps with a stable-ish link for discussion etc.) and deal with more exchanges in the subsequent edit PRs.

Adding of `while!` in this PR may never return depending the condition @jwosty is meeting at each evaluation.

PR #474, opened 2020-06-18
Record resolutions for open static classes

dsyme:


dsyme:
See discussion in https://github.com/fsharp/fslang-design/issues/352#issuecomment-646243221 after design review with @tihan and @cartermp

PR #475, opened 2020-06-25
Add design principles for string interpolation

dsyme:
Based on conversation with @TIHan


dsyme:
Anyone interested in this feature should check out these design principles

PR #477, opened 2020-06-27
Add design review meeting to FS-1001 string interpolation

dsyme:
Yesterday @cartermp, @tihan, @jonsequitur and I met for a design review meeting for string interpolation.  @cartermp made notes.

I propose that for ever RFC we add a checkbox "Design review meeting" which must get a checkmark to proceed the RFC, and link to the notes and attendees.

thanks
don

### DIFF
--- a/RFCs/FS-1001-StringInterpolation.md
+++ b/RFCs/FS-1001-StringInterpolation.md
@@ -3,9 +3,9 @@
 
 There is an approved-in-principle [proposal](http://fslang.uservoice.com/forums/245727-f-language/suggestions/6002107-steal-nice-println-syntax-from-swift) to extend the existing printf functionality in the F# language design with string interpolation. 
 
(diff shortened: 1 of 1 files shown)


abelbraaksma:
That sounds great for the general RFC process, I assume it'll apply to RFCs that have been "approved in principle", and are about to be implemented, or already have an (unfinished) implementation.

The notes, will they be made available (like they do with the runtime team)?

cartermp:
The notes in this case were captured in the PR by both @dsyme and myself.

As far as a general process goes - I think for extremely simple designs we shouldn't require a meeting. But for something like this, where it affects the very core of working with strings, we need a design review face-to-face.

dsyme:
> As far as a general process goes - I think for extremely simple designs we shouldn't require a meeting. But for something like this, where it affects the very core of working with strings, we need a design review face-to-face.

Yes agreed

PR #480, opened 2020-06-28
RFC-1090 for unmanaged generic struct types

vzarytovskii:
An RFC for [generic struct type whose fields are all unmanaged types is unmanaged](https://github.com/fsharp/fslang-suggestions/issues/692) suggestion.

### DIFF
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -122,7 +122,7 @@ type MyStructGenericWithUnusedUnmanagedParameter<'T>(x: 'T, y: 'T, z: obj) =
     member _.X = x
     member _.Y = y
 
-// Not unmanaged, since it has a field with a non-unmanaged type.
+// Not unmanaged, since it has a field with a managed type.
 [<Struct>]
 type MyStructGenericWithUnmanagedField<'T>(x: 'T, y: 'T, z: obj) =
     member _.X = x
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -69,8 +69,6 @@ stdin(6,13): error FS0001: A generic construct requires that the type 'MyStructG
 
 This proposal aims to resolve this inconsistency by treating any generic struct type as `unmanaged` if all its fields are unmanaged.
 
-__NOTE:__ Both generic struct types _with_ and _without_ the `unmanaged` constraint on type parameter(s) can be treated as unmanaged as long as type parameters are being substituted by an unmanaged type(s).
-
 For example, consider the following code:
 
 ```fsharp
@@ -86,6 +84,8 @@ type MyStructGenericWithNoConstraint<'T>(x: 'T, y: 'T) =
 ```
 
 Instances of both `MyStructGeneric<'T>` and `MyStructGenericWithNoConstraint<'T>` will be treated as unmanaged type as long as `'T` is being unmanaged.
+
+In other words, any generic struct type, with all of its fielads are known to be unmanaged, can be considered unmanaged, _with_ or _without_ the `unmanaged` constraint on type parameter(s)
 
 ## Adjusted definition of an unmanaged type
 
@@ -116,7 +116,7 @@ type MyStructGenericWithNoConstraint<'T>(x: 'T, y: 'T) =
     member _.Y = y
 
 // Can be considerd unmanaged, as long as 'T is unmanaged.
-// Note, that despite constructor having and unmanaged argument (obj), it is not used as part of the backing field.
+// Note, that despite constructor having managed argument (obj), it is not used as part of the backing field.
 [<Struct>]
 type MyStructGenericWithUnusedUnmanagedParameter<'T>(x: 'T, y: 'T, z: obj) =
     member _.X = x
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -11,7 +11,7 @@ This RFC covers the detailed proposal for this suggestion.
 
 [summary]: #summary
 
-Allow generic structs to be `unmanaged` at construction if all fields are `unmanaged` types, so they can be used where `unmanaged` constraint is applied.
+Allow generic structs to be `unmanaged` at construction if all fields are `unmanaged` types.
 
 # Motivation
 
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -109,7 +109,7 @@ type MyStructGeneric<'T when 'T: unmanaged>(x: 'T, y: 'T) =
     member _.X = x
     member _.Y = y
 
-// Will be unmanaged, asl long as 'T is unmanaged.
+// Can be considered unmanaged, as long as 'T is unmanaged.
 [<Struct>]
 type MyStructGenericWithNoConstraint<'T>(x: 'T, y: 'T) =
     member _.X = x
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -53,7 +53,7 @@ let error = Test<MyStructGeneric<int>>()
 Prior to this RFC, the example above will fail to compile with:
 
 ```less
-let error = Test<Test<int>>()
+let error = Test<MyStructGeneric<int>>()
   ------------^^^^^^^^^^^^^^^
 
 stdin(6,13): error FS0001: A generic construct requires that the type 'Test<int>' is an unmanaged type
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -26,7 +26,7 @@ Allow generic structs to be `unmanaged` at construction if all fields are `unman
 
 Currently, it is not possible to have an `unmanaged` constraint working on generic struct types.
 
-Consider following code:
+Consider the following code:
 
 ```fsharp
 [<Struct>]
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -56,7 +56,7 @@ Prior to this RFC, the example above will fail to compile with:
 let error = Test<MyStructGeneric<int>>()
   ------------^^^^^^^^^^^^^^^
 
-stdin(6,13): error FS0001: A generic construct requires that the type 'Test<int>' is an unmanaged type
+stdin(6,13): error FS0001: A generic construct requires that the type 'MyStructGeneric<int>' is an unmanaged type
 ```
 
 This proposal aims to eliminate this restriction by treating generic struct type `unmanaged` if all its fields are `unmanaged`.
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -24,7 +24,7 @@ Allow generic structs to be `unmanaged` at construction if all fields are `unman
 
 [design]: #detailed-design
 
-Currently, it is not possible to have an `unmanaged` constraint on generic struct types.
+Currently, it is not possible to have an `unmanaged` constraint working on generic struct types.
 
 Consider following code:
 
@@ -41,12 +41,13 @@ let error = Test<Test<int>>()
 Prior to this RFC, the example above will fail to compile with:
 
 ```less
-stdin(27,1): error FS0001: A generic construct requires that the type 'Test<int>' is an `unmanaged` type
+let error = Test<Test<int>>()
+  ------------^^^^^^^^^^^^^^^
+
+stdin(6,13): error FS0001: A generic construct requires that the type 'Test<int>' is an unmanaged type
 ```
 
 This proposal aims to eliminate this restriction by treating generic struct type `unmanaged` if all its fields are `unmanaged`.
-
-The
 
 ## Examples
 
--- a/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
+++ b/RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
@@ -17,7 +17,7 @@ Allow generic structs to be `unmanaged` at construction if all fields are `unman
 
 [motivation]: #motivation
 
-* Make it easier to author low level interop code.
+* Make it easier to author low-level code.
 * Improve interoperability with C#.
 
 # Detailed design
@@ -33,21 +33,29 @@ Consider the following code:
 type MyStruct(x: int, y: int) =
     member _.X = x
     member _.Y = y
-    
+
 [<Struct>]
 type MyStructGeneric<'T when 'T: unmanaged>(x: 'T, y: 'T) =
     member _.X = x
     member _.Y = y
 
 [<Struct>]
+type MyStructGenericWithNoConstraint<'T>(x: 'T, y: 'T) =
+    member _.X = x
+    member _.Y = y
+
+[<Struct>]
 type Test<'T when 'T: unmanaged> =
     val element: 'T
-        
+
 let works = Test<int>()
 let works2 = Test<MyStruct>()
 
 // error FS0001: A generic construct requires that the type 'MyStructGeneric<int>' is an unmanaged type
 let error = Test<MyStructGeneric<int>>()
+
+// error FS0001: A generic construct requires that the type 'MyStructGenericWithNoConstraint<int>' is an unmanaged type
+let error = Test<MyStructGenericWithNoConstraint<int>>()
 ```
 
 Prior to this RFC, the example above will fail to compile with:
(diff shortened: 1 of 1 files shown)


RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,130 @@
...
+[summary]: #summary
+
+Allow generic structs to be `unmanaged` at construction if all fields are `unmanaged` types.
+
+# Motivation
+
+[motivation]: #motivation
+
+* Feature parity with C#.
+* Improve interoperability with C#.
```
cartermp, line 21: Feature party with C# is not a goal in and of itself. Interop is, so that should stay.

The chief motivation here should be from the F# perspective: better support for low-level programming in F#, specifically when it comes to offer provably safe compile-time analysis so that we don't allocate needlessly. It's actually nearly identical to the [C# motivation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.3/blittable), though there may be some F#-specific considerations here.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,130 @@
...
+
```
cartermp, line 22: With two goals - better low-level F# programming and better C# interop - there should be two examples of code you'd like to write here that you can't write in F# today. One about doing low-level programming entirely in F#, and another using a C# API that assumes (in some form) that a caller is unmanaged (but only if you cannot use that API today).

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,130 @@
...
+Prior to this RFC, the example above will fail to compile with:
+
+```less
+let error = Test<Test<int>>()
+  ------------^^^^^^^^^^^^^^^
+
+stdin(6,13): error FS0001: A generic construct requires that the type 'Test<int>' is an unmanaged type
+```
+
+This proposal aims to eliminate this restriction by treating generic struct type `unmanaged` if all its fields are `unmanaged`.
```
cartermp, line 50: This is the sort of thing that should go in the motivation section.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,130 @@
...
+
```
cartermp, line 51: Here we'll need a few things:

* What the definition of an unmanaged type is
* How the existing [`umanaged`](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/generics/constraints) constraint is to be adjusted (if at all)
* All types that are considered unmanaged (e.g., the numeric primitives)
* Example of custom struct that is unmanaged and one that is not
* How do we consider struct records? Allowed, subject to specific rules, disallowed, etc.
* How do we consider struct anonymous records? Allowed, subject to specific rules, disallowed, etc.
* How do we consider struct single-case DUs? Allowed, subject to specific rules, disallowed, etc.
* How do we consider struct multi-case DUs? Allowed, subject to specific rules, disallowed, etc.
* How do we deal with the `modreq` on existing APIs marked with them for this constraint?
* Do we also emit an `modreq`?
* C# has their [own metadata representation ](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.3/blittable#metadata-representation)since the F# constraint put them in a little design hole. What do we do about that?

(Note on `modreq` - we don't really use it correctly. @dsyme and @TIHan can fill you in on that bag of fun)

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,130 @@
...
+## Examples
```
cartermp, line 52: These are great from the F# perspective. Pending consideration for records and DUs, we'll also want examples for that. We may also want to show what it's like interoperating with a .NET API marked with the constraint.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,130 @@
...
+    test Unchecked.defaultof<S<int>>
+
+let not_passing () =
+    test Unchecked.defaultof<S<obj>>
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+No drawbacks.
```
cartermp, line 120: I suppose there is technically is a drawback insofar as it has a limited F# audience, since most F# programmers are not doing low-level programming. But it is somewhat of a "completeness" feature that fits in with other work done in F# 4.5 to make F# more suitable for low-level work.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: I've updated examples, supported types, posted a description of the unmanaged types, etc.
I will need some time to figure out `modreq` part though.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: Updated

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: Added some description of supported types, based on the current implementation PR.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: Updated

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+[summary]: #summary
+
+Allow generic structs to be `unmanaged` at construction if all fields are `unmanaged` types.
+
+# Motivation
+
+[motivation]: #motivation
+
+* Make it easier to author low level interop code.
```
cartermp, line 20: Although this will primarily be for calling APIs that likely do some native interop, it could be possible to be 100% F# code. So I think this is probably better as "low-level code", since that encapsulates both.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+* Improve interoperability with C#.
+
+# Detailed design
+
+[design]: #detailed-design
+
+Currently, it is not possible to have an `unmanaged` constraint working on generic struct types.
+
+Consider following code:
```
cartermp, line 29:
```suggestion
Consider the following code:
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+```fsharp
+[<Struct>]
+type Test<'T when 'T: unmanaged> =
+        val element: 'T
+
+let works = Test<int>()
+let error = Test<Test<int>>()
+
```
cartermp, line 38: I recommend the code sample also includes an example of a custom struct that does work.

```suggestion
[<Struct>]
type MyStruct(x: int, y: int) =
    member _.X = x
    member _.Y = y
    
[<Struct>]
type MyStructGeneric<'T when 'T: unmanaged>(x: 'T, y: 'T) =
    member _.X = x
    member _.Y = y

[<Struct>]
type Test<'T when 'T: unmanaged> =
    val element: 'T
        
let works = Test<int>()
let works2 = Test<MyStruct>()

// error FS0001: A generic construct requires that the type 'MyStructGeneric<int>' is an unmanaged type
let error = Test<MyStructGeneric<int>>()
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+[motivation]: #motivation
+
+* Make it easier to author low level interop code.
+* Improve interoperability with C#.
+
+# Detailed design
+
+[design]: #detailed-design
+
+Currently, it is not possible to have an `unmanaged` constraint working on generic struct types.
```
cartermp, line 27: This will need additional clarification:

```suggestion
Currently, it is not possible to define a generic struct with the `unmanaged` constraint and substitute the type with another generic struct, even if its type is also constrained to be `unmanaged`.
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+let works = Test<int>()
+let error = Test<Test<int>>()
+
+```
+
+Prior to this RFC, the example above will fail to compile with:
+
+```less
+let error = Test<Test<int>>()
```
cartermp, line 44:
```suggestion
let error = Test<MyStructGeneric<int>>()
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+  ------------^^^^^^^^^^^^^^^
+
+stdin(6,13): error FS0001: A generic construct requires that the type 'Test<int>' is an unmanaged type
```
cartermp, line 47:
```suggestion
stdin(6,13): error FS0001: A generic construct requires that the type 'MyStructGeneric<int>' is an unmanaged type
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+This proposal aims to eliminate this restriction by treating generic struct type `unmanaged` if all its fields are `unmanaged`.
```
cartermp, line 50:
```suggestion
This proposal aims to resolve this inconsistency by treating a generic struct type as `unmanaged` if all its fields are unmanaged.
```

cartermp: The  way this is worded makes it sound like I won't need to specify `unmanaged` in a generic struct definition, and it will "just work" so long as the `'T` being substituted is unamanaged. Is that correct, or do I need to be explicit in my type definition?

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Definition
```
cartermp, line 52:
```suggestion
## Adjusted definition of an unmanaged type
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
cartermp: This also means that this won't work either:

```fsharp
[<Struct>]
type MyStruct(x: byref<int>, y: byref<int>) =
    member _.X = x
    member _.Y = y
```

Which feels like a big gap in the feature area.

cartermp: Nor this

```fsharp
[<Struct>]
type MyStruct(x: Span<int>, y: Span<int>) =
    member _.X = x
    member _.Y = y
```

Nor this:

```fsharp
let fails = Test<Span<int>>()
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+An unmanaged-type is any type that isn't a reference-type and contains no fields whose type is not an unmanaged-type.
```
cartermp, line 54: In this section it would be good to lay out the existing F# definition and then give the adjustment, like so:

```suggestion
The existing definition of an `unamanged` type is:

* `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
* Any enum type.
* Any `nativeptr<_>`.
* Any non-generic user-defined struct-type that can be statically determined to be 'unmanaged' at construction.
```

We adjust this definition by modifying the last section:

```
Any user-defined struct type that can be statically determined to be 'unmanaged' at construction.
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+In other words, an unmanaged-type is one of the following:
+
+* `sbyte, byte, short, ushort, int, uint, long, ulong, char, float, double, decimal, or bool`.
+* Any enum-type.
+* Any pointer-type.
+* Any generic user-defined struct-type that can be statically determined to be 'unmanaged' at construction.
```
cartermp, line 61: This section would be obviated with the suggested layout.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
```
cartermp, line 62: Would be good to give a simple example of a user-defined struct that is unmanaged vs. one that is not unmanaged.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+let test (x: 'T when 'T : unmanaged) = ()
+
+let passing () =
+    test Unchecked.defaultof<S<int>>
+
+let not_passing () =
+    test Unchecked.defaultof<S<obj>>
+```
+
+Union types:
```
cartermp, line 147: Needs examples where it doesn't work

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+```fsharp
+[<Struct>]
+type Container<'a> = Container of 'a
+
+let test (x: 'T when 'T : unmanaged) = ()
+
+test (Container 1)
+```
+
+Tuples:
```
cartermp, line 158: Needs example where it doesn't work

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+
+```fsharp
+let x = struct(1,2,3);;
+
+let test (x: 'T when 'T : unmanaged) = ()
+
+test (x)
+```
+
+Records:
```
cartermp, line 168: Needs example where it doesn't work

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+An unmanaged-type is any type that isn't a reference-type and contains no fields whose type is not an unmanaged-type.
+
+In other words, an unmanaged-type is one of the following:
+
+* `sbyte, byte, short, ushort, int, uint, long, ulong, char, float, double, decimal, or bool`.
+* Any enum-type.
+* Any pointer-type.
+* Any generic user-defined struct-type that can be statically determined to be 'unmanaged' at construction.
+
+## Implementation details
```
cartermp, line 63: Needs some more info:

* Can I mix `unamnaged` with other generic constraints? Right now I can actually write the following code that looks like absolute nonsense.

```fsharp
[<Struct>]
type Test<'T when 'T: unmanaged and 'T: not struct> =
    val element: 'T
```

This and other silly things you can do today should be corrected as a part of this change. `unmanaged` implies `struct` in C#, which I would expect to be the case here. We'll need an example of doing this and what the error message should say.

* Clarify that the constraint is enforced by the language, not the runtime

* Clarify that nested fields in structs must also be unmanaged, with examples of working/not-working

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,198 @@
...
+## Supported types
+
+* Generic struct types.
+* Struct records.
+* Anonymous struct records.
+* Struct tuples.
+  * Struct tuples are thread as unmanaged as long as all the elements are unmanaged;
+* Struct unions.
+  * Single-case unions types are treated as unmanaged as long as underlying type is unmanaged.
+  * Multi-case unions are treated as unmanaged as long as all underlying types are unmanaged.
```
cartermp, line 78:
```suggestion
* Generic struct types whose fields are unmanaged
* Struct records with unmanaged fields
* Anonymous struct records with unmanaged fields
* Struct tuples with unmanaged values
* Struct unions (single and multi-case) whose fields are all unmanaged
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: > * Can I mix `unamnaged` with other generic constraints? Right now I can actually write the following code that looks like absolute nonsense.

That is actually one of the things I wanted to add, and didn't, because I assumed it works already (I thought that you cannot mix 'unmanaged' and 'not struct' now).
I will add a point for fixing this.
Thanks for noticing.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: @cartermp Yeah, the idea is that you won't need to specify the unmanaged constraint and it will "just work".
I will add a note about it in the RFC.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: Added a point to implementation details.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,288 @@
...
+    member _.X = x
+    member _.Y = y
+
+[<Struct>]
+type MyStructGenericWithNoConstraint<'T>(x: 'T, y: 'T) =
+    member _.X = x
+    member _.Y = y
+```
+
+Instances of both `MyStructGeneric<'T>` and `MyStructGenericWithNoConstraint<'T>` will be treated as unmanaged type as long as `'T` is being unmanaged.
```
cartermp, line 88: I think I understand this section, but it's a little confusing (perhaps due to my feedback).

Both of these struct types should be usable as an unmanaged type. That is, I could do this:

```fsharp
let t = Test<MyStructGeneric<int>>()
let t' = Test<MyStructGenericWithNoConstraint<int>>()
```

Having the constraint on the type I am substituting the `'T` in `Test` with is not required with this change.

In other words, any generic struct where all its fields are known to be unmanaged can be considered unmanaged.

Kind of a mouthful and getting the definition across clearly is kind of difficult. But I think if you include examples of instantiating `Test` with both it will be clearer.

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,288 @@
...
+Examples:
+
+```fsharp
+// Always unmanaged (forced by constraint).
+[<Struct>]
+type MyStructGeneric<'T when 'T: unmanaged>(x: 'T, y: 'T) =
+    member _.X = x
+    member _.Y = y
+
+// Will be unmanaged, asl long as 'T is unmanaged.
```
cartermp, line 112:
```suggestion
// Can be considered unmanaged, as long as 'T is unmanaged.
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,288 @@
...
+[<Struct>]
+type MyStructGenericWithNoConstraint<'T>(x: 'T, y: 'T) =
+    member _.X = x
+    member _.Y = y
+
+// Will be unmanaged, asl long as 'T is unmanaged.
```
cartermp, line 118:
```suggestion
// Can be considerd unmanaged, as long as 'T is unmanaged.
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,288 @@
...
+// Note, that despite constructor having and unmanaged argument (obj), it is not used as part of the backing field.
```
cartermp, line 119: > unmanaged argument (obj)

Do you mean managed?

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,288 @@
...
+[<Struct>]
+type MyStructGenericWithUnusedUnmanagedParameter<'T>(x: 'T, y: 'T, z: obj) =
+    member _.X = x
+    member _.Y = y
+
+// Not unmanaged, since it has a field with a non-unmanaged type.
```
cartermp, line 125:
```suggestion
// Not unmanaged, since it has a field with a managed type.
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
```diff
@@ -0,0 +1,288 @@
...
+
+* Generic struct types whose fields are unmanaged
+* Struct records with unmanaged fields
+* Anonymous struct records with unmanaged fields
+* Struct tuples with unmanaged values
+* Struct unions (single and multi-case) whose fields are all unmanaged
+
+## Examples
+
+The following examples are not possible now and will be once feature is implemented.
```
cartermp, line 151:
```suggestion
The following examples are now possible with this feature:
```

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: Yeah, typo, will fix

RFCs/FS-1090-Generic-struct-type-whose-fields-are-all-unmanaged-types-is-unmanaged.md
vzarytovskii: I've changed some wording, hope it's more clear now :)

dsyme:
LGTM

PR #481, opened 2020-07-02
Update and rename FS-1068-open-static-classes.md to FS-1068-open-type…

TIHan:
…-declaration.md


preview/FS-1068-open-type-declaration.md
```diff
@@ -186,84 +175,46 @@ N(12.0) // Works!
...
-        }
-    }
-}
-```
+See original discussion here: https://github.com/fsharp/fslang-design/issues/352#issuecomment-499146012.  
 
 #### Combining overloaded methods from different types
 
-Post-preview we will allow combination of method overloads from different types according to C# rules.
+We will **not** allow combination of method overloads from different types according to C# rules. The reasoning is for interactivity; we need to be able to shadow method groups in order to not constantly have old method overloads exist.
```
cartermp:
```suggestion
We will **not** allow combination of method overloads from different types according to C# rules. The reasoning is for interactivity scenarios such as FSI or Jupyter Notebooks: we need to be able to shadow method groups in order to not constantly have old method overloads exist.
```

preview/FS-1068-open-type-declaration.md
```diff
@@ -186,84 +175,46 @@ N(12.0) // Works!
...
-        {
-            var v1 = new Vector3();
-            var result1 = Lerp(v1, v1, 1);
-            var v2 = new Vector2();
-            var result2 = Lerp(v2, v2, 1);
-        }
-    }
-}
-```
+See original discussion here: https://github.com/fsharp/fslang-design/issues/352#issuecomment-499146012.
```
cartermp, line 188: This can just be deleted now, since the updated design will ship as preview, thus obviating this whole section

preview/FS-1068-open-type-declaration.md
```diff
@@ -36,43 +34,26 @@ Type providers can provide static classes, hence this would allow type providers
...
-type C =
+type A() =
     static member M(x: int) = x * 2
 
-open type C
+open type A
 
-M(2) |> ignore
+M(2)
 ```
```
cartermp, line 50: Needs explanation of the syntax `open type` and how this is a new kind of open declaration.

preview/FS-1068-open-type-declaration.md
```diff
@@ -291,33 +242,6 @@ Try to resolve `member-ident` to one of the following, in order:
...
-```
-
-This appears barely documented in the C# design docs. The preview version of the F# feature explicitly does not allow this: `open type` is only allowed on non-generic static classes.
-
-From an interop perspective, opening a generic static class may be required as some APIs may use generic type parameters on the static class, forcing users to open these with a concrete substitution when using them.
-
-We will not do this for F#.
-
-
 ## Compatibility
```
cartermp, line 245: Should have a section on tooling compatibility here. What are implications in tooling for this change, and what needs/needed to be done to ensure a good experience?

preview/FS-1068-open-type-declaration.md
TIHan: Was the summary not sufficient?

TIHan:
@cartermp @dsyme Anything need to be updated here? Any comments?

preview/FS-1068-open-type-declaration.md
cartermp: There's no formal description of the new syntax form. We'll need that, similar to what was done with `nameof`.

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+
+* [x] Approved in principle
+* [x] Details: [Resolved to Preview](https://github.com/fsharp/fslang-design/issues/352)
+* [x] Implementation: [Complete to Preview](https://github.com/dotnet/fsharp/pull/6325)
+
+
+## Summary
+[summary]: #summary
+
+Add support for opening types to gain access to static members and nested types, e.g. `open type System.Math`. For example
```
cartermp, line 17: This isn't defining what opening a type is. It explains what it can do, but not what it is. We're adding a new kind of open declaration to the language - `open type SomeType`, which allows for unqualified access to its static members and nested types. Once a definition like this is in place then this summary is good.

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+> Min(2.0, 1.0);;
+val it : float = 1.0
+```
+
+## Motivation
+[motivation]: #motivation
+
+This greatly increases the expressivity of F# DSLs by allowing method-API facilities such as named arguments, optional arguments and type-directed overloading to be used in the DSL design.
+
+Type providers can provide static classes, hence this would allow type providers to provide "unqualified" names. This useful for some projections where the natural thing is unqualified, e.g. the "R" type provider always required `R.plot` etc. With this feature, `R.plot` can now be `plot`.
```
cartermp, line 33:
```suggestion
Type providers can provide static classes, hence this would allow type providers to provide "unqualified" names. This is useful for some projections where the natural thing is unqualified, e.g. the "R" type provider always required `R.plot` etc. With this feature, `R.plot` can now be `plot`.
```

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+
+Additionally, important C# DSL APIs are starting to appear that effectively require this.
+
+## Detailed design
+[design]: #detailed-design
+
+Opening of types allows for treating a type somewhat as if it were a module with module-bound functions inside of it. For example:
```
cartermp, line 40: This is a little too vague. We need to be specific about what is exposed - static members and nested types - and then follow up by saying it is similar to module-bound functions.

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+```fsharp
+type A() =
+    static member M(x: int) = x * 2
+
+open type A
+
+M(2)
+```
+
+However, **members are not functions**, so very different rules apply beyond simple cases like the above. This primarily concerns name resolution when overloads are involved, or when considering whether or not we resolve a method or a property first.
```
cartermp, line 51: > However, **members are not functions**, so very different rules apply beyond simple cases like the above. This primarily concerns name resolution when overloads are involved, or when considering whether or not we resolve a method or a property first.

Neither of these are related to module-bound functions as far as I can tell. You make mention of rules, including the application of some rules in the previous example, but there is no mention of rules for resolving either a module-bound function or a member.

Additionally, what are the concerns? I don't see any description about how we resolve overloads or the resolution of a name of a method or a property. There is an alternatives section describing the construction of a method overload group, but what does it contrast with?

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+    static member M(x: int) = x * 2
+
+open type C // Compile error
+
+M(2) // 'M' is not recognized
+```
+
+#### `AutoOpen`
+
+Also like modules, specifying `AutoOpen` on a type automatically brings its members into scope.
```
cartermp, line 89: Needs mention that the exposed _things_ (not just members) are subject to the rules of name resolution that need defining above.

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+[<AutoOpen>]
+type A() =
+    static member M(x: int) = x * 2
+
+M(2) // Can call 'M' without opening 'A'
+```
+
+### Opening types reveals nested types
+
+When opening a .NET type (or a provided type), any accessible nested types also become accessible without qualification.
```
cartermp, line 101: How do you open a provided type? Can there be an example of getting access to the nested type?

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+
+open type C
+
+M(2)
+N(2.0) // Works!
+```
+
+### Opening types with generic parameter instantiations
+
+You can open types with generic parameter instantiations:
```
cartermp, line 148: Can you open it multiple times with different instantiations? e.g.

```fsharp
open type Vector<int>
open type Vector<float>
```

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+open type (int * int) // Compile error, named types are only allowed
+```
+
+Using an abbreviation will still not allow it:
+```fsharp
+type MyAbbrev = (int * int)
+open type MyAbbrev // Compile error, named types are only allowed
+```
+
+These rules are very similar when using `inherit` in a type definition.
```
cartermp, line 177:
```suggestion
These rules are very similar to when you use `inherit` in a type definition.
```

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+### Named Types
+
+Named, or nominal, types are only allowed. It means you cannot open a function, tuple, or anonymous record:
+```fsharp
+open type (int * int) // Compile error, named types are only allowed
+```
+
+Using an abbreviation will still not allow it:
```
cartermp, line 171:
```suggestion
Using an abbreviation will also not allow it:
```

preview/FS-1068-open-type-declaration.md
```diff
@@ -0,0 +1,273 @@
...
+1. A union case.
+2. **A method group.**
+3. A property group.
+4. A field.
+5. An event.
+6. **A method group of extension members, by consulting the `ExtensionsInScope` table.**
+7. A property group of extension members , by consulting the `ExtensionsInScope` table.
+8. A nested type `type-nested`. Recursively resolve .rest if it is present, otherwise return `type-nested`
+
+That is to say, we will resolve methods over properties in the context of opening static classes or static members extending a static class.
```
cartermp, line 261: This section reads as if we're going to do this. Should either remove it or reword it to say that it's the avenue we _would_ take.

dsyme:
@TIHan @cartermp Would you like me to make the revisions to @cartermp 's excellent comments or would you like to iterate on it between you?  Thanks

TIHan:
You can make revisions @dsyme unless there is something specific we should talk.

preview/FS-1068-open-type-declaration.md
TIHan: You can, but the latest one would shadow all nested types/method groups from the first one.

preview/FS-1068-open-type-declaration.md
dsyme: I adjusted the test.  It's not very formal still but I think it's what we need for now

dsyme:
OK I made the revisions, we can iterate from here, thanks!

dsyme:
FUther things we need to clarify, i'' add these to the doc too:

1. Noticed by Phillip

    ```
    open type Vector<int>
    open type Vector<float>
    ```


2. inherited members.  I notice in the current preview that **any** `open System.Math` means that the static members of `System.Object` come into scope so `Equal` and `ReferenceEqual` are is now available. We should match C# here

TIHan:
> inherited members. I notice in the current preview that any open System.Math means that the static members of System.Object come into scope so Equal and ReferenceEqual are is now available. We should match C# here

Glad we will be explicit about it; we already made the changes for this in the PR implementation. Will add test cases to ensure we don't bring it in scope.

dsyme:
> Glad we will be explicit about it; we already made the changes for this in the PR implementation. Will add test cases to ensure we don't bring it in scope.

Cool

PR #486, opened 2020-07-14
Update FS-1068-open-type-declaration.md

dsyme:
Adds two cases mentioned here:

https://github.com/fsharp/fslang-design/pull/481#issuecomment-658303028

### DIFF
--- a/preview/FS-1068-open-type-declaration.md
+++ b/preview/FS-1068-open-type-declaration.md
@@ -181,6 +181,23 @@ open System.Numerics
 open type Vector<_> // Compile error
 ```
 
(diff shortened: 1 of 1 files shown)


dsyme:
@TIHan Looking over the C# specification https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-static I also see we need to make a modification for extension members.

Here's the important text I've added so far:


### Inherited members

`open type` only makes accessible static members and nested types declared in the specified type. Inherited members are not imported.

### C# Extension members

C#-style extension members are declared as static members with the `[<Extension>]` attribute.

`open type` makes extension methods declared in the specified type available for extension method lookup.
However, the names of the extension methods are not imported into scope for unqualified reference in code.

PR #495, opened 2020-08-22
RFC 1091 - Extending the Types which Support Units of Measure

pblasucci:
RFC 1091 - Extending the Types which Support Units of Measure

### DIFF
--- a/RFCs/FS-1091-Extend-Units-of-Measure.md
+++ b/RFCs/FS-1091-Extend-Units-of-Measure.md
@@ -8,7 +8,7 @@ This RFC covers the detailed proposal for this suggestion.
 - [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/901)
 - [x] [Implementation](https://github.com/dotnet/fsharp/pull/9978)
 - [ ] Design Review Meeting(s) with @dsyme and others invitees
-- [ ] [Discussion](https://github.com/fsharp/fslang-design/issues/PLEASE-ADD-A-DISCUSSION-ISSUE-AND-LINK-HERE)_(n.b. pending the acceptance of this RFC)_
+- [x] [Discussion](https://github.com/fsharp/fslang-design/issues/496)
 
 # Summary
 
--- a/RFCs/FS-1091-Extend-Units-of-Measure.md
+++ b/RFCs/FS-1091-Extend-Units-of-Measure.md
@@ -1,0 +1,142 @@
+# F# RFC FS-1091 - Extend Units of Measure to Include More Numeric Types
+
+The design suggestion [Extend Units of Measure to Include Unsigned Integers](https://github.com/fsharp/fslang-suggestions/issues/901) has been marked "approved in principle".
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] Approved in principle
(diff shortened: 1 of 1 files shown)


abelbraaksma:
> In terms of level of effort, the items marked MAYBE? in the preceding table are no more difficult than the items marked YES (in the same column). Though this was equally true back when signed integers were added. So I just wanted to make sure there wasn't some theoretical or practical (or philosophical) reason of which I'm not aware. :-)

I can't answer that last question, but for parity and orthogonality, given that it's trivial, I feel we should implement this. After all, we do it for a given set of type aliases, I feel it's odd we don't do it for the others.

Which reminds me, I haven't even checked if it worked on the types themselves, can you do `let f (x: Int32<kg>) = 42`?

abelbraaksma:
> Units of measure can be applied to any type, not just floating point types; however, only floating point types, signed integral types, and decimal types support dimensioned quantities. Therefore, it only makes sense to use units of measure on the primitive types and on aggregates that contain these primitive types.

That's a quote from the docs: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/units-of-measure. I'm not sure how to read it, because the first sentence appears to be inaccurate: I don't see a (trivial) way to use UoM on any type.

pblasucci:
I'm in complete agreement with you @abelbraaksma re: consistency across abbreviations. 💯

I've also not checked if it's supported on the _actual_ type (😱). But based on my current understanding of the code, I doubt it.

abelbraaksma:
> But based on my current understanding of the code, I doubt it.

Indeed, it doesn't. We replied on the same moment, but see my quote above.

pblasucci:
I think the quote is alluding to `MeasureAnnotatedAbbreviationAttribute` and/or generic measures, both of which can be applied for any (custom) type. But I'm not certain.

abelbraaksma:
I wonder, though, if the current type system (once you make your changes) would consider these distinct types, or correct aliases:

```f#
let f (x: int<kg>) (y: Int32<kg>): int32<kg> = x + y
let g (x: int<kg>) (y: Int32<kg>): int32<kg^2> = x *  y
```

pblasucci:
🤔 This will lead to some fun experiments when I'm back at my laptop later tonight... but I'd expect the current (released) compiler to fail 2 and 3 as invalid. Should be easy-enough to correct, though.

abelbraaksma:
Currently they are invalid, indeed, what I'm curious about is whether they're considered invalid after you'd have made the changes to allow more aliases, and the actual type, to have UoM.

pblasucci:
Ah, I see... I investigated this exact scenario last night.  :frowning: 

I still believe it's possible. But it will be more work than I originally thought. It seems some additional logic needs to be injected somewhere (probably the constraint solver), so that measured abbreviations are considered in terms of their base types (without sacrificing the measure information).

Alternatively, we could:

+ use only a single abbreviation for measured values (i.e. same as today)
+ have the various abbreviations be unequal to one another (low effort, but a _violates the principle of least surprise_)

'course, I've only spent an hour or two on this. So I think it's worth digging in a bit further.

abelbraaksma:
I think we ought to open a discussion thread, which appears to be customary for RFC's that require some discussion. I opened one here: https://github.com/fsharp/fslang-design/issues/496. You may add this as a link in the RFC for future reference.

PR #499, opened 2020-09-04
Add RFC-1092 for support of System.Half

abelbraaksma:
See language suggestion: https://github.com/fsharp/fslang-suggestions/issues/909 and discussion for this RFC: https://github.com/fsharp/fslang-design/issues/500.

### DIFF
--- a/RFCs/RFC-1092-support-cli-half-type-as-float16.md
+++ b/RFCs/RFC-1092-support-cli-half-type-as-float16.md
@@ -8,7 +8,7 @@ This RFC covers the detailed proposal for this suggestion.
 - [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/909)
 - [ ] [Implementation] tbd
 - [x] Design Review Meeting(s) with @dsyme and others invitees: n/a
-- [x] [Discussion](https://github.com/fsharp/fslang-design/issues/TBD)
+- [x] [Discussion](https://github.com/fsharp/fslang-design/issues/500)
 
 # Summary
 
--- a/RFCs/RFC-1092-support-cli-half-type-as-float16.md
+++ b/RFCs/RFC-1092-support-cli-half-type-as-float16.md
@@ -1,0 +1,77 @@
+# F# RFC FS-1092 - Support CLI half type as float16
+
+The design suggestion [Support new CLI type: half type/float16](https://github.com/fsharp/fslang-suggestions/issues/909) has been marked "approved in principle".
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] Approved in principle
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/909)
+- [ ] [Implementation] tbd
+- [x] Design Review Meeting(s) with @dsyme and others invitees: n/a
+- [x] [Discussion](https://github.com/fsharp/fslang-design/issues/TBD)
+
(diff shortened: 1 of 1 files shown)


RFCs/RFC-1092-support-cli-half-type-as-float16.md
```diff
@@ -0,0 +1,77 @@
...
+  * No
+* What happens when previous versions of the F# compiler encounter this design addition as source code?
+  * They continue to work
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+  * Unless both binaries are compiled against .NET 5.0, this may lead to assembly loading errors.
+* If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct?
+  * We should take proper measures to allow FSharp.Core to be linked with other versions of .NET that do not have this type.
+
+
+# Unresolved questions
```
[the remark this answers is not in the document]
abelbraaksma: I agree, I put that in the discussion thread too. For this and future dependencies on features added to net core or NetStandard, we need perhaps some bespoke way to implement that in FSharp.Core so that it can still be used interchangeably with older versions too.

RFCs/RFC-1092-support-cli-half-type-as-float16.md
```diff
@@ -0,0 +1,77 @@
...
+// x is of type `half`
+let x = half 34.56
+
+// x is of type `half`
+let x = 42.99h
+```
+
+# Drawbacks
+
+If we don't do this, people could still use `System.Half` directly, but don't have good language support, like overloaded conversion functions and proper compilation to IL.
```
dsyme, line 53: I would like to understand the current experience of using System.Half and see a comparison of what does and doesn't work when the user simply defines an alias `type half = System.Half`.

My belief is the experience won't be that bad but some conversions will be missing, as will a `half` function.

I don't know of anything that falls under the "proper compilation to IL" category.

We should be precise about the current experience.

dsyme:
Will need to switch to RFCS-1094 or something due to erased unions grabbing the number

kerams:
One more thing for the design section - `%f` in format strings will need to accept `Half`.

RFCs/RFC-1092-support-cli-half-type-as-float16.md
abelbraaksma: @dsyme It's been a while, but the current experience is indeed not too bad. Though users do not have formatters in `sprintf` and the like, nor do they have any ability for creating constants. The latter might actually be made possible if we were to allow constants of any blittable type (???).

I doubt `System.Half` is widely used presently. It's main use case is in CUDA and AI programming.

While I'd like this feature (and `Int128`, for that matter), unless we find a way to support features from .NET 6.0 and up, I don't quite see this happening: https://github.com/fsharp/fslang-design/discussions/500#discussioncomment-9459078.

T-Gro:
With the introduction of https://github.com/dotnet/fsharp/pull/20229, we can now have if-def-based additions to FSharp.Core that rely on newer runtime APIs. We could therefore reference `System.Half` (net5+) under `#if NET5_0_OR_GREATER` for the `float16` alias, the `half` conversion, and the generic-math wiring in `prim-types.fs` — shipped only in the net asset, leaving the netstandard builds untouched. (Only a `1.0h` literal would additionally need lexer support.)

PR #507, opened 2020-10-26
Create FST-1033-analyzers.md - Towards F# Analyzer support in FSharp.Compiler.Service

dsyme:
This adds an RFC discussing what to do about getting F# Analyzer support in to the heart of F# tooling, so it lights up everywhere

It's just a discussion initially, basically about binary compatibility and opening up the question whether we could first start with analyzers which internalise their own parsing/analysis.

cc @Krzysztof-Cieslak @cartermp @forki @baronfel


dsyme:
I'll merge this as it's easier to read in situ, and add a discussion thread.

PR #512, opened 2020-12-06
Add RFC-1092: Support for erased union types

Swoorup:
As requested by @dsyme on twitter. 

This adds an RFC for erased union types. First time here writing an RFC so might not have fully thought things through. 

The suggestion: https://github.com/fsharp/fslang-suggestions/issues/538 probably needs to be approved first. If everything is fine, happy to continue with the actual implementation.

[Rendered](https://github.com/swoorup/fslang-design/blob/master/RFCs/FS-1092-Erased-Union-Types.md)

### DIFF
--- a/RFCs/FS-1092-Erased-Union-Types.md
+++ b/RFCs/FS-1092-Erased-Union-Types.md
@@ -1,4 +1,4 @@
-# F# RFC FS-1092 - Erased Union Types
+﻿# F# RFC FS-1092 - Erased Union Types
 
 This RFC covers the detailed proposal for this suggestion. [Erased type-tagged anonymous union types](https://github.com/fsharp/fslang-suggestions/issues/538).
 
@@ -87,7 +87,7 @@ However the following is valid:
 
 ```fsharp
 // inferred to (int|string)
-let intOrString = if true then 1 :> (int|string) else "Hello" :> _ 
+let intOrString = if true then 1 :> (int|string) else "Hello" :> _
 ```
 
 This respects the rules around where explicit upcasting is required including cases despite where type information being available. Although the later might change depending on the outcome of [fslang-suggestion#849](https://github.com/fsharp/fslang-suggestions/issues/849)
@@ -101,6 +101,15 @@ If the selector of a pattern match is an erased union type, the match is conside
 let prettyPrint (x: (int8|int16|int64|string)) =
     match x with
     | :? (int8|int16|int64) as y -> prettyPrintNumber y
+    | :? string as y -> prettyPrintNumber y
+```
+
+The above is the same as F# in current form:
+
+```fsharp
+let prettyPrint (x: obj) =
+    match x with
+    | :? int8 | :? int16 | :? int64 as y -> prettyPrintNumber y
     | :? string as y -> prettyPrintNumber y
 ```
 
@@ -120,17 +129,38 @@ The IL wrapping type for `(A | B)` is the _smallest intersection type_ of base
 types of `A` and `B`. For example:
 
 ```fsharp
-type IntOrString = (int|string) // wrapping type is System.Object
-type IntOrString = (int8|int16|float) // wrapping type is System.ValueType
+// wrapping type is System.Object
+type IntOrString = (int|string)
+// wrapping type is System.ValueType
+type IntOrString = (int8|int16|float)
 type I = interface end
 type A = inherit I
 type B = inherit I
-type AorB = (A|B) // I is the wrapping type
+// I is the wrapping type
+type AorB = (A|B) 
 
 type I2 = interface end
 type C = inherit I inherit I2
 type D = inherit I inherit I2
-type CorD = (C|D) // Both I or I2 could be potential wrapping type. The compiler would choose I2 since its the earliest ancestor
+// Both I or I2 could be potential wrapping type. The compiler would choose I2 since its the earliest ancestor
+type CorD = (C|D) 
+```
+
+## Hierarchies in Types
+
+```
+   ┌───┐
+   │ I │
+   └─┬─┘
+  ┌──┴───┐
+┌─┴─┐  ┌─┴─┐
+│ A │  │ B │
+└───┘  └─┬─┘
+      ┌──┴───┐
+    ┌─┴─┐  ┌─┴─┐
+    │ C │  │ D │
+    └───┘  └───┘
+
 ```
 
 # Drawbacks
--- a/RFCs/FS-1092-Erased-Union-Types.md
+++ b/RFCs/FS-1092-Erased-Union-Types.md
@@ -39,8 +39,8 @@ type WorkManagerMessage = (RunWork | SubscribeProgressUpdate)
 
 let processWorkerMessage (msg: WorkerMessage) =
     match msg with
-    | :? RunWork -> ...
-    | :? RequestProgressUpdate -> ...
+    | :? RunWork as m -> ...
+    | :? RequestProgressUpdate m -> ...
 ```
 
 ```fsharp
@@ -127,8 +127,21 @@ TBD
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-* Initial implementation should not allow for using uom in erased unions?
+
+* Initial implementation should not allow for using uom in erased unions? 
+
+    ```fsharp
+    type [<Measure>] userid
+    type UserId = int<userid>
+    type IntOrUserId = (int|UserId)
+    ```
+
 * Initial implementation should not allow using generic type arguments in erased unions?
+
+    ```fsharp
+    type StringOr<'a> = ('a | string)
+    ```
+
 * Initial implementation should not allow for common members of the erased unions to be exposed without upcasting?
 
     ```fsharp
@@ -150,3 +163,4 @@ TBD
     let what = shape.What // error
     let what = (shape :> IShape).What // ok
     ```
+* Should exhaustive check in instance check be implemented across normal circumstances? https://github.com/dotnet/fsharp/issues/10615
--- a/RFCs/FS-1092-Erased-Union-Types.md
+++ b/RFCs/FS-1092-Erased-Union-Types.md
@@ -1,0 +1,152 @@
+# F# RFC FS-1092 - Erased Union Types
+
+This RFC covers the detailed proposal for this suggestion. [Erased type-tagged anonymous union types](https://github.com/fsharp/fslang-suggestions/issues/538).
+
+* [ ] Approved in principle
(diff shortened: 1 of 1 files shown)


RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+Add erased union types as a feature to F#. Erased Union types provide some of the benefits of structural ("duck") typing, within the confines of a nominative type system.
+
+# Motivation
+[motivation]: #motivation
+
+Supporting erased union types in the language allows us to move more type information with the usual advantages this brings:
+
+* They serve as an alternative to function overloading.
+* They obey subtyping rules.
+* They allow representing subset of protocols as a type without needing to resort to the lowest common denominator like `obj`.
```
cartermp, line 23: DUs allow this today, too. What is the benefit of erased/ad-hoc unions as opposed to existing ones for this point? Perhaps this is worth expounding upon in an example later on.

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+* Types are actually enforced, so mistakes can be caught early.
```
cartermp, line 24: Same comment re: DUs

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+* They allow representing more than one type
```
cartermp, line 25: Same comment re: DUs

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+* Because they are enforced, type information is less likely to become outdated or miss edge-cases.
```
cartermp, line 26: Same comment re: DUs

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+type Username = Username of string
+type Password = Password of string
+type UserOrPass = (Password | UserName) // UserOrPass is a type alias
+
+// `getUserOrPass` is inferred to `unit -> UserOrPass`
+let getUserOrPass () = if (true) then name :> UserOrPass else password :> UserOrPass
+
+// `getUserOrPass2` binding is inferred to `unit -> (UserOrPass | Error)`
+let getUserOrPass2 () = if isErr() then err :> (UserOrPass | Error) else getUserOrPass() :> _
+```
```
cartermp, line 56: These examples are fine, but they don't appear to build on any of the motivation. I'd recommend just having a simple code example here that demonstrates "what the feature is all about" and then dive into more examples in appropriate sub-sections.

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+* Erased union are commutative and associative:
+
+    ```fsharp
+    (A | B) =:= (B | A)
+    (A | (B | C)) =:= (( A | B ) | C)
+    ```
+
+    *`=:=` implies type equality and interchangable in all context*
+
+* If `A :> C` and `B :> C` then `(A | B) :> C` where `T :> U` implicies T is subtype of C;
```
cartermp, line 73:
```suggestion
* If `A :> C` and `B :> C` then `(A | B) :> C` where `T :> U` implies T is subtype of C;
```

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+
+type (A|B|I) // equal to type definition for I, since I is supertype of A and B
+type (A|B|C) // equal to type (A|B), since B is supertype of C
+type (A|C)   // disjoint as A and C both inherit from I but do not have relationship between each other.
+```
+
+## Type inference
+[inference]: #type-inference
+
+Erased Union type is explicitly inferred meaning that at least one of the type in an expression must contain the erased union type.
```
cartermp, line 105:
```suggestion
Erased Union type is explicitly inferred meaning that at least one of the types in an expression must contain the erased union type.
```

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+However the following is valid:
+
+```fsharp
+// inferred to (int|string)
+let intOrString = if true then 1 :> (int|string) else "Hello" :> _
+```
+
+This respects the rules around where explicit upcasting is required including cases despite where type information being available. Although the later might change depending on the outcome of [fslang-suggestion#849](https://github.com/fsharp/fslang-suggestions/issues/849)
```
cartermp, line 120:
```suggestion
This respects the rules around where explicit upcasting is required including cases despite where type information being available. Although the latter might change depending on the outcome of [fslang-suggestion#849](https://github.com/fsharp/fslang-suggestions/issues/849)
```

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+
+    ```fsharp
+    type [<Measure>] userid
+    type UserId = int<userid>
+    type IntOrUserId = (int|UserId)
+    ```
+
+    Alternatively we could just warn when such constructs are used.
+
+* Initial implementation should not allow using static or generic type arguments in erased unions?
```
cartermp, line 199: It should definitely support generics in the first version

Swoorup: @cartermp Things like 
```fsharp
let a: list<(string|int>) = ...
``` 
would work. 

But I am not sure how generics parameter in union can even be supported? As it is erased, the wrapping type cannot be known when creating IL for generics? My knowledge in this area is limited. Am I missing something?

Would you expect to write in form of 
```fsharp
let formUnion<'a, 'b> (test: 'a): ('a|'b) = test :> _
```

or 
```fsharp
let formUnion<'a, 'b when 'a :> ('a |'b) and 'b :> ('a | 'b)> (test: 'a) = test :> _
```

kerams: I suppose the most obvious solution is using the closest common ancestor as the generic type behind the scenes. That would often be `obj`, causing boxing for any value types among the constituent types. On the other hand, given this type

```fsharp
// wrapping type is System.Object
type IntOrString = (int|string)
```

`int`s that you pass around this way will need to be boxed as well.

However, if you were to use a proper union to represent `IntOrString`, you'd still need to allocate... Perhaps boxing should be seen as an unovoidable fact?

Swoorup: That would be foregoing where Erased type is `ValueType` in situations where if all cases are structs. But F# appears to do box and unbox anyway in case struct is upcasted to ValueType.

kerams: Yeah, it does https://sharplab.io/#v2:DYLgZgzgPsCmAuACAhiRBlAnhesC2AdAGrLACusAKpgA6wDaAuogLyL1SICMA3IgEwREURkA
It makes sense when you think about it -- individual value types can have various sizes, how would you pack them in an array?

Swoorup: That makes sense. Only situation I imagine where this might be bit overhead is when reduction of joins of same type or join of subtype to supertype leads to same type. For example: `(int|int)` or `(IA | I)` where `I :> IA`. But union case are only special cases though

RFCs/FS-1092-Erased-Union-Types.md
matthewcrews: Am I allowed to add examples? I didn't want to step on anyone's toes.

dsyme: Either add a comment with your examples, or send a pull request to the pull request branch  :)

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,253 @@
...
+
+Becomes
+
+``` fsharp
+type Decision =
+    // Fields
+    
+    abstract member (*) (a: (float|Decision), b:Decision) : LinearExpression =
+        match a with
+        | :> float as f -> // float action
```
dsyme, line 78: why not `:?` instead of `:>`

matthewcrews: No reason. That is the more correct thing to use in this situation. I was referencing some other code that I had done which relied on casting. I would prefer `:?`.

matthewcrews: The big thing for me is easing reasoning and long term support of domains with A) many operator overloads and B) a more robust form of method/function overloading. Two major pain points I would much rather have this than function overloading. I believe it is a cleaner and more sustainable approach.

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,253 @@
...
+        | :> Decision as d -> // Decision action
+```
+
+The maintenance of libraries with large numbers of operator-overloads becomes simpler because the behavior is defined in one place.
+
+# Detailed design
+[design]: #detailed-design
```
dsyme, line 85: In this section you should first describe the syntax added (for types and expressions and patterns).

Then describe what happens when these are checked, e.g. what makes these valid

* are type variables allowed `('T | int)`
* would `(int | int)` give a warning

RFCs/FS-1092-Erased-Union-Types.md
dsyme: @cartermp I'm inclined to say that this feature should not support naked generic type variables, e.g. `(int | 'T)`.  This is just not needed to get the utility of this feature and brings in a huge range of issues which we can't solve (e.g. what happens on substitution for T).  If you want naked generic type variables you use a tagged union

dsyme:
I'm going to merge this to get a linkable RFC then we can address the feedback above in a separate PR

RFCs/FS-1092-Erased-Union-Types.md
```diff
@@ -0,0 +1,227 @@
...
+[summary]: #summary
+
+Add erased union types as a feature to F#. Erased Union types provide some of the benefits of structural ("duck") typing, within the confines of a nominative type system.
+
+# Motivation
+[motivation]: #motivation
+
+Supporting erased union types in the language allows us to move more type information with the usual advantages this brings:
+
+* They serve as an alternative to function overloading.
```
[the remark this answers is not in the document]
isaacabraham: Or to method overloading.

PR #515, opened 2021-01-13
Add RFC for `Map.maxKeyValue` and `Map.minKeyValue` functions

sangkilc:
See the [language suggestion](https://github.com/fsharp/fslang-suggestions/issues/933).

### DIFF
--- a/RFCs/FS-1092-map-min-max-keyvalue.md
+++ b/RFCs/FS-1092-map-min-max-keyvalue.md
@@ -50,18 +50,24 @@ or `Map.fold`, which requires iterating the whole entries in the worst case.
 
 # Detailed design
 
-I didn't look at the actual implementation of `Map` yet, but it should be
-straightforward to implement these functions. One can implement this by simply
-traversing the internal tree (`O(log n)`). For example, in order to find the
-minimum key-value pair, one can recursively follow the "left" of the tree.
+Since the `Map` is implementing the traditional self-balancing persistent tree,
+one can implement this by simply traversing the internal tree. The complexity
+should be `O(log n)`. For example, in order to find the minimum key-value pair,
+one can recursively follow the "left" of the tree.
 
-We may consider adding `tryMinKeyValue` and `tryMaxKeyValue`, too.
+In the case where the given map is empty, both the functions should raise
+`KeyNotFoundException` as in
+https://github.com/dotnet/fsharp/blob/main/src/fsharp/FSharp.Core/map.fs#L158.
+
+To gracefully handle exceptions without `try-with`, we may consider adding
+`tryMinKeyValue` and `tryMaxKeyValue`, too. These functions are only useful when
+the given map is empty, so this is not absolutely necessary.
 
 Example code:
 
 ```fsharp
-let x = Map.tryMaxKeyValue 42 db // Return an option
 let x = Map.maxKeyValue 42 db // Return the result or raise an exception
+let x = Map.maxKeyValue 42 Map.empty // This will raise an exception
 ```
 
 # Drawbacks
--- a/RFCs/FS-1092-map-min-max-keyvalue.md
+++ b/RFCs/FS-1092-map-min-max-keyvalue.md
@@ -1,0 +1,87 @@
+# F# RFC FS-1092- Add Map.maxKeyValue and Map.minKeyValue functions
+
+The design suggestion [Add Map.maxBinding and Map.minBinding functions](https://github.com/fsharp/fslang-suggestions/issues/933) has been marked "approved in principle".
+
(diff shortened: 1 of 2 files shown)


RFCs/FS-1092-map-min-max-keyvalue.md
```diff
@@ -0,0 +1,87 @@
...
+  findLargestAge db |> printfn "%d"
+```
+
+In this case, we would like to find out the largest age in the
+database. Currently the only way to get the largest age is to use `Map.filter`
+or `Map.fold`, which requires iterating the whole entries in the worst case.
+
+# Detailed design
+
+I didn't look at the actual implementation of `Map` yet, but it should be
```
cartermp, line 53: For the purposes of the document, I recommend looking into the implementation and then writing it from the perspective of:

"The implementation is `O(log n)` by doing [...]"

Or something to that effect. Since maps are currently backed by an AVL tree that would be true.

RFCs/FS-1092-map-min-max-keyvalue.md
```diff
@@ -0,0 +1,87 @@
...
+minimum key-value pair, one can recursively follow the "left" of the tree.
+
+We may consider adding `tryMinKeyValue` and `tryMaxKeyValue`, too.
+
+Example code:
+
+```fsharp
+let x = Map.tryMaxKeyValue 42 db // Return an option
+let x = Map.maxKeyValue 42 db // Return the result or raise an exception
+```
```
cartermp, line 65: What would be your rational for having option-returning functions as well? The only case that I think is relevant here would be the empty map, and I'm not sure it's worth adding two extra functions for that.

RFCs/FS-1092-map-min-max-keyvalue.md
```diff
@@ -0,0 +1,87 @@
...
+  let db = addMember 42 "Charlie"
+  (* Find the largest age in the db *)
+  findLargestAge db |> printfn "%d"
+```
+
+In this case, we would like to find out the largest age in the
+database. Currently the only way to get the largest age is to use `Map.filter`
+or `Map.fold`, which requires iterating the whole entries in the worst case.
+
+# Detailed design
```
cartermp, line 51: What would be the behavior in the case of the empty map? What if there is only one element? We'll need this specified as well.

RFCs/FS-1092-map-min-max-keyvalue.md
sangkilc: It was true. I modified the description accordingly.

RFCs/FS-1092-map-min-max-keyvalue.md
sangkilc: It is not absolutely necessary. I modified the description, too.

RFCs/FS-1092-map-min-max-keyvalue.md
sangkilc: Fixed too.

cartermp:
@dsyme could you also take a look at this design proposal?

RFCs/FS-1092-map-min-max-keyvalue.md
```diff
@@ -0,0 +1,93 @@
...
+- [x] Approved in principle
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/933)
+- [ ] [Implementation](https://github.com/dotnet/fsharp/pull/FILL-ME-IN)
+- [ ] Design Review Meeting(s) with @dsyme and others invitees
+- [Discussion](https://github.com/fsharp/fslang-design/issues/PLEASE-ADD-A-DISCUSSION-ISSUE-AND-LINK-HERE)
+
+# Summary
+
+This feature extends the `Map` module by adding two functions, tentatively named
+`Map.maxKeyValue` and `Map.minKeyValue`. Those functions allow users to look up
```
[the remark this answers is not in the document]
krauthaufen: Why not simply add both?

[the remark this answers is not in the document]
sangkilc: I would really appreciate if you consider adding `Map.maxBinding` or `Map.maxKeyValue` or similar to eliminate a second lookup.

RFCs/FS-1092-map-min-max-keyvalue.md
```diff
@@ -0,0 +1,93 @@
+# F# RFC FS-1092- Add Map.maxKeyValue and Map.minKeyValue functions
```
dsyme, line 1: new number will be needed

sangkilc: Fixed

PR #520, opened 2021-01-20
Update FS-1092-Erased-Union-Types.md

dsyme:
@Swoorup I've gone over the RFC and adjusted it

The motivation section has been re-written.  I'm open to including more motivating cases though in each case it should be explained why using a labelled discriminated union has problems.  That's always the baseline comparison point.

I'm de-emphasizing anything to do with subtyping or hierarchies.  These are important technical issues but we do not want this mechanism to be an entry for normalising hierarchical classification as a commonly used technique in F# (since it is almost always used badly outside a few well known use-cases).

### DIFF
--- a/RFCs/FS-1092-Erased-Union-Types.md
+++ b/RFCs/FS-1092-Erased-Union-Types.md
@@ -10,51 +10,56 @@ This RFC covers the detailed proposal for this suggestion. [Erased type-tagged a
 # Summary
 [summary]: #summary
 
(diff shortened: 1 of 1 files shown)


dsyme:
I made some subsequent edits by direct commit.

@Swoorup Could you explain more the practical use cases that are making you want this feature?  

I'd also like more use case from Javascript

Swoorup:
@dsyme 
I couldn't have explain it better than @Horusiath in the context of actor and type safe-error channels.

https://github.com/fsharp/fslang-design/discussions/519#discussioncomment-295606

PR #521, opened 2021-01-20
FS-1092: Minor fixes 

Swoorup:
Just few minor changes (grammar and few sample code).

> The mechanism can encourage "hierarchy thinking" where the user wastes precious thought time on trying to find a perfect "classification" of disparate cases into a set of hierarchically organised types. This kind of activity is normally unproductive, leading to fragile code and "false" attempts at finding commonality.

Would have thought this to be a motivation, not drawback :P


dsyme:
Thanks for the cleanup!

Issue #410, opened 2019-10-26
RFC FS-1046 is assigned twice


inosik:
While looking for the specs for the "Wildcard self identifiers" feature, I noticed that FS-1046 was already assigned before, for ["Consistent tuple types"](https://github.com/fsharp/fslang-design/blob/6229e6048ad49ec2b1820f5211703c4bf6220730/FSharp-4.1b/FS-1046-consistent-tuple-types.md).

dsyme:
We'll put this down as a quirk of history

PR #536, opened 2021-03-30
Draft RFC for [<RequireNamedArgument>]

smoothdeveloper:
See
* suggestion: https://github.com/fsharp/fslang-suggestions/issues/414
* discussion: https://github.com/fsharp/fslang-design/discussions/538

### DIFF
--- a/drafts/FS-1095-requirenamedargumentattribute.md
+++ b/drafts/FS-1095-requirenamedargumentattribute.md
@@ -1,0 +1,84 @@
+# F# RFC FS-NNNN - [<RequireNamedArgument>] attribute
+
+The design suggestion [Add an attribute enforcing the use of named argument at callsite
+](https://github.com/fsharp/fslang-suggestions/issues/414) has been marked "approved in principle".
+
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] Approved in principle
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
Ready for review and merge to avoid someone taking my first RFC id 🙂

Happypig375:
As if non-official RFCs get merged https://github.com/fsharp/fslang-design/pull/442

smoothdeveloper:
@Happypig375 I'm not following what you mean, is there an extra step to get the document merged?

I'm reading on the readme.md that a PR should be submitted to "grab the token"

> Ideas which get "approved in principle" get an RFC entry based on the template, and a corresponding RFC discussion thread

But if it doesn't get merged to enact the "RFC entry", then #410 will be a reoccurring issue.

If the RFC draft (it is by no mean a final document since the discussion didn't take place) isn't merged, it is hard to have the discussion on it, etc.

Are you saying this shouldn't get merged? if so, should I add a warning about the RFC being "non-official" at the very top of the text?

If your concern holds, the readme.md needs clarification under "The Process".

Happypig375:
I'm just saying that you can't really expect quick PR merging here.

smoothdeveloper:
My inference is not as good as F# type checker, thanks for the additional context 🙂

PR #545, opened 2021-04-27
Split tasks out to separate RFC

dsyme:
This splits everything to do with tasks out into a separate RFC


Happypig375:
Where is taskSeq?

dsyme:
> Where is taskSeq?

RIght now it's in the `tests/fsharp/perf/tasks/FS` folder, so there's no RFC for it yet, it's just a sample of what you can do with resumable code.   One reason I split out the RFCs was to write separate RFCs for taskSeq etc.

PR #562, opened 2021-05-21
Update FS-1097-task-builder.md

dsyme:
I managed to greatly simplify the use of SRTP for tasks.

As explanation, the use of `CanBind` and `CanReturnFrom` multi-parameter constraints with witnesses had been brought over from TaskBuilder.fs.  However it seems it is no longer needed - either because of the removal of context-insensitive tasks, or else because of some fixes in constraint processing, or perhaps the low/medium/high priority technique for overload resoltuion now being used.


dsyme:
The corresponding commit in the implementation is here: https://github.com/dotnet/fsharp/pull/6811/commits/6a7c42f5372cefa82197ca99069befd88ea6f15b

PR #570, opened 2021-05-26
Implementation proposed for FS-1100 - Printf format as binary number

Happypig375:
How do I initiate "Design Review Meeting(s) with @dsyme and others invitees"?


smoothdeveloper:
Unsure, but I'd say https://en.wiktionary.org/wiki/Hollywood_principle (without the implication that they won't call).

It is an `async` code block and you've just made it hot, but you can't make it `RunSynchronously` 🙂

cartermp:
In this case it won't be necessary, since the change is small enough in scope that the design and rationale is quite straightforward. Compare that with tasks, which needs _multiple_ sessions 😄 . Feel free to delete that section and/or mark it as checked.

cartermp (approved):

PR #571, opened 2021-05-26
Update FS-1097-task-builder.md

dsyme:
Update spec of `backgroundTask { ... }`, based on https://github.com/rspeele/TaskBuilder.fs/issues/35#issuecomment-848077410

### DIFF
--- a/RFCs/FS-1097-task-builder.md
+++ b/RFCs/FS-1097-task-builder.md
@@ -95,10 +95,14 @@ Tasks are executed immediately to their first await point. For example:
         }
     printfn "x = %d" x // prints "x = 1"
 ```
-Background tasks (ignoring any SynchronizationContext, starting in thread pool)
+Background tasks are specified using `backgroundTask { ... }`.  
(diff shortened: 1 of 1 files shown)


NinoFloris:
I think it shouldn't do this, backgroundTask screams 'everything inside of this will not run on this thread' yet a single `while true` at the start would now halt my UI.

Even if we'd decide to keep this new semantics just checking SyncCtx is not enough. Tasks also respect `TaskScheduler.Current` meaning we must include another check to see if it's anything but `TaskScheduler.Default`.

smoothdeveloper:
@dsyme, AFAIR, this is a split of a larger RFC, but it still points to the larger RFC and I think the semantics of `task` from end user point of view don't depend on the compiler machinery of the larger RFC.

Would it make sense to have a separate new discussion thread on just `task` CE?

dsyme:
> Would it make sense to have a separate new discussion thread on just `task` CE?

It would

PR #576, opened 2021-05-30
RequireNamedArgumentAttribute: minor adjustments

smoothdeveloper:
Added a couple of unresolved questions:
* delegate/function pointers support
* type checking performance consideration

Also, resolve some questions with most likely decision based on the discussions.


cartermp (approved):
Thanks!

PR #586, opened 2021-06-03
Update FS-1039-struct-representation-for-active-patterns.md

dsyme:
Trim down this RFC to only cover the part about struct returns for active patterns


dsyme:
I did a direct commit of RFC FS-1104 to cover the other parts which were factored out in this PR

PR #591, opened 2021-06-08
Update FS-1092-anonymous-type-tagged-unions.md

charlesroddie:
Mention the need to avoid exposing usage of this feature to non-F# .Net languages.

### DIFF
--- a/RFCs/FS-1092-anonymous-type-tagged-unions.md
+++ b/RFCs/FS-1092-anonymous-type-tagged-unions.md
@@ -88,6 +88,8 @@ An anonymous type-tagged union type should only be considered for when a union i
 
 5. There is some identified, concrete, simply explained benefit over using labelled discriminated unions, e.g. "we have a simpler API with fewer overloads".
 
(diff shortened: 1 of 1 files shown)


dsyme (approved):

RFCs/FS-1092-anonymous-type-tagged-unions.md
```diff
@@ -88,6 +88,8 @@ An anonymous type-tagged union type should only be considered for when a union i
 
 5. There is some identified, concrete, simply explained benefit over using labelled discriminated unions, e.g. "we have a simpler API with fewer overloads".
 
+6. There is no possibility that the type will end up being consumed by .Net languages other than F#.
```
dsyme, line 91: This is a bit strong - it might be reasonable to expose - but I know what you mean :)

cartermp: Similar thoughts, but it's a good thing to clarify for now. Maybe as the feature takes shape a bit more we'll have a more nuanced position

PR #592, opened 2021-06-08
Update FS-1093-additional-conversions.md

dsyme:
I'm removing the `int32` --> `float32` and `single` --> `double` type-directed conversions from the proposal. The motivations for these are just too weak to justify the disturbance in the force for how F# type checking works.  

This leaves just the following additions for numeric widenings:

    int32 --> int64
    int32 --> float64

These have strong use cases as documented in the RFC (e.g. the first because int64 is now so often the standard for integer data)


dsyme:
@vzarytovskii I thought it over after our discussion and I do think it's better to be more conservative on this.

PR #595, opened 2021-06-09
Create FS-1105-Non-variable-patterns-to-the-right-of-as-patterns.md

Happypig375:


drafts/FS-1105-Non-variable-patterns-to-the-right-of-as-patterns.md
```diff
@@ -0,0 +1,66 @@
...
+| :? struct(int * int) as (x, y) -> printfn $"A struct tuple: {x}, {y}"
+| :? DU as DU (x, y) -> printfn $"A DU: {x}, {y}"
+| :? int as Id i -> printfn $"An int: {i}"
+| _ -> printfn "Nope"
+```
+are now valid.
+
+It will act similarly as the `&` operator requiring both sides of the operator to have a successful match before succeeding.
+
+Although [the `as` operator has the lowest precedence of all oeprators](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/#operator-precedence), the right hand side of `as` operators currently only parse one identifier, which means that `,`, `|`, `&`, `::`, `:`, and `as` operators can still appear to the right of identifiers after `as` operators and be regarded as valid code. Therefore, the right side of `as` operators should parse a pattern with an equivalent precedence to DU case matching and active patterns, which means that patterns with lower precedence will require parentheses.
```
smoothdeveloper, line 38:
```suggestion
Although [the `as` operator has the lowest precedence of all operators](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/#operator-precedence), the right hand side of `as` operators currently only parse one identifier, which means that `,`, `|`, `&`, `::`, `:`, and `as` operators can still appear to the right of identifiers after `as` operators and be regarded as valid code. Therefore, the right side of `as` operators should parse a pattern with an equivalent precedence to DU case matching and active patterns, which means that patterns with lower precedence will require parentheses.
```

Swoorup:
Might be easier for others to review if you add something like a **Rendered** hyperlink that links to your RFC file in your branch. 

Like this: https://github.com/rust-lang/rfcs/pull/3137

Swoorup:
@Happypig375 Might be easier for others to review if you add something like a **Rendered** hyperlink that links to your RFC file in your branch. 

Like this: https://github.com/rust-lang/rfcs/pull/3137

Happypig375:
Is it that hard to click "Files changed" -> "..." -> "View file"? [Rendered](https://github.com/Happypig375/fslang-design/blob/patch-17/drafts/FS-1105-Non-variable-patterns-to-the-right-of-as-patterns.md)

Swoorup:
Up until now, I had no idea you could do that.

Happypig375:
There is also the opportunity to make `1 | 2 as :? int` happen. This will save one level of parentheses over `:? int as (1 | 2)`. What are your thoughts?

dsyme:
> There is also the opportunity to make 1 | 2 as :? int happen. This will save one level of parentheses over :? int as (1 | 2). What are your thoughts?

I don't think this should be included. The pattern on the left matches type "int", the pattern on the right matches type "obj".   We should tighten typing on a left-to right basis, so yes to `:? int as 3` and not `3 as :? int`

PR #605, opened 2021-07-08
Remove excluded conversions in FS-1093

Happypig375:


dsyme (approved):

dsyme:
FWIW I think we should also include `int32` --> `nativeint`

PR #630, opened 2021-09-08
Document the `id` attribute for `example` nodes

baronfel:

### DIFF
--- a/tooling/FST-1031-xmldoc-extensions.md
+++ b/tooling/FST-1031-xmldoc-extensions.md
@@ -171,6 +171,50 @@ with a language identifier, but this practice is widespread in popular community
 This attribute often allows for language-specific highlighting or other user experience benefits and so we recommend adding it
 whenever possible.
 
+For example:
+
+```fsharp
+namespace MyLibrary.Core
+
+    open System
+
+    type MyType() = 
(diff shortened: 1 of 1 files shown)


tooling/FST-1031-xmldoc-extensions.md
```diff
@@ -171,6 +171,50 @@ with a language identifier, but this practice is widespread in popular community
...
+       ///</example>
+       member x.P = 1
+```
+    
+
+## `<example id="<custom id>">`
+
+The `example` XML documentation element is not specified to accept an `id` attribute for purposes of uniquely identifying an
+example, but doing so allows tooling to deep-link to a specific example and so should be encouraged. `id` attributes _should_
+be url-safe by default, and _should_ be unique across a library.
```
baronfel, line 197: @dsyme question about guidance here: what level of uniqueness scope is enough? Should tools ensure uniqueness via namespacing, or should users ensure uniqueness via naming on their own?

dsyme: fsdocs isn't exactly great about diagnostics and error failures

I guess it should be "warn if strict is not on, error if it is"

dsyme (approved):

PR #651, opened 2021-12-02
Add info about style guide management

nojaf:
Discussing with @dsyme in person, we've agreed that the issues arising about the F# style guide are better managed via the authority of the F# design process.  Hence we've transferred eight issues into this repo, and marked them, and this updates the README appropriately


nojaf:
@dsyme temporarily made me admin here so we can merge this directly

PR #658, opened 2022-03-26
RFC FS-1088: fix code example

3Rafal:
`else!` isn't a construct proposed in this RFC, so I assume it should be replaced with `elif!`.


cartermp (approved):
Thanks, good catch!

PR #659, opened 2022-04-21
Update FST-1027-fsi-references.md

smoothdeveloper:
Just clarifying the status on "Handler Resolution", for now.

### DIFF
--- a/tooling/FST-1027-fsi-references.md
+++ b/tooling/FST-1027-fsi-references.md
@@ -160,7 +160,11 @@ We may consider using `AssemblyLoadContext` in some clever way in the future, bu
 
 It works in all FCS driven tool (or even fsc itself) through :
 * having the extension assembly next to the process one
-* the `--compilertool:`  flag, whilch can point to a directory containing the extensions
+* the `--compilertool:` flag, whilch can point to a directory containing the extensions
+
+By convention, tools that load extensions in other locations, should also load extensions that are sitting in same folder, unless there are good reasons not doing so.
+
+By convention, tools that load extensions, should never fail to do so for binding redirect matters related to FSharp.Core version the extensions must have been compiled against.
 
 Note: additional stable locations were considered by not implemented in earlier drafts of this RFC.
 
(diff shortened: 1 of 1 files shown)


tooling/FST-1027-fsi-references.md
```diff
@@ -156,7 +156,19 @@ This will generate an error because `System.Whatever` version `4.1.0` will alrea
...
+
+It works in all FCS driven tool (or even fsc itself) through :
+* having the extension assembly next to the process one
+* the `--compilertool:` flag, whilch can point to a directory containing the extensions
+
+By convention, tools that load extensions in other locations, should also load extensions that are sitting in same folder, unless there are good reasons not doing so.
+
+By convention, tools that load extensions, should never fail to do so for binding redirect matters related to FSharp.Core version the extensions must have been compiled against.
+
+Note: additional stable locations were considered by not implemented in earlier drafts of this RFC.
```
dsyme, line 169: `by` --> `but`

Could you check the whole of the rest of the RFC and see if there are incorrect things that should be removed or moved to an "Alternatives" section please? Thanks

smoothdeveloper: @dsyme, I've made few more edits on what seems inaccurate; but for whole which remains in Runtime section, I am not sure if it matches the reality and I didn't author those details.

It is likely some of those aspects of the RFC were done before/during "how does this play with dotnet core and how reference assemblies for netstandard stuff work with FSI" period.

For the RFC, I think it is @KevinRansom and @cartermp that can help farther narrow down this particular RFC if needed.

Some general remarks about RFCs, maybe they should only give overview from end user perspective; aside, we could have a postmortem document of RFC that made it to production, focused on summarising key aspects of the implementation details and considerations (here, it would be a documentation on the reflection bits as they are used to make the extension system work).

smoothdeveloper:
@dsyme I am happy with update, it takes the RFC closer to the actual situation for handler resolution, and preserves the suggested approach, which hasn't been mulled over / implemented.

Main concern for me is still https://github.com/dotnet/fsharp/issues/12895

dsyme:
Thanks!

PR #667, opened 2022-06-11
Add Result module parity to the option module RFC

uxsoft:


vzarytovskii (approved):

vzarytovskii (approved):

cartermp (approved):

cartermp:
Not required for the RFC, but this will also require extensive API `///` comments just like the Option module.

uxsoft:
> Not required for the RFC, but this will also require extensive API `///` comments just like the Option module.

Here's the implementation incl. examples. I stripped out the examples here for brewity.

PR #674, opened 2022-06-22
Add FS-1125-print-println-functions.md

bisen2:
Language suggestion: https://github.com/fsharp/fslang-suggestions/issues/1092
Discussion: https://github.com/fsharp/fslang-design/discussions/675


RFCs/FS-1125-print-println-functions.md
```diff
@@ -0,0 +1,69 @@
...
+println str // prints "A string" with a trailing newline
+println $"The value in str is %s{str}" // prints "The value in str is A string" with a trailing newline
+println $"%0.3f{System.Math.PI}" // prints "3.142" with a trailing newline
+```
+
+# Drawbacks
+
+The addition of these functions brings the core library to four different `print*` functions that are automatically available and an even higher number of `*print*` functions. There is some additional cognitive load to differentiating between these functions, particularly for experienced F# users who are already accustomed to `printf` and `printfn`.
+
+# Alternatives
```
cartermp, line 48: I believe `printn` was discussed as an option, might be good to list here just for the sake of completeness, even if we're probably not gonna go with `printn` as an option

cartermp (approved):
Thanks! I'll approve for now and then leave open for any additional review for a time, and if there's no further questions will merge.

cartermp:
Merging, thanks!

PR #681, opened 2022-06-30
Update FS-1124-interfaces-with-static-abstract-members.md

smoothdeveloper:
minor suggestions for review, I'd also like if

> Do not mix static and non-static interfaces in IWSAMs.

could be expanded upon to provide the whys.


dsyme:
Hmmmm I get what you're saying but I think these changes aren't clarifying

Will close and keep in mind

PR #698, opened 2022-07-21
Improve signature of defaultWith

edgarfgp:

Based on @gusty suggestion to improve the defaultWith signature : 
See : https://github.com/fsharp/fslang-suggestions/issues/1123#issuecomment-1179171067
- Will give us the opportunity to use the error value in the compensation.

### DIFF
--- a/RFCs/FS-1123-result-module-parity-with-option.md
+++ b/RFCs/FS-1123-result-module-parity-with-option.md
@@ -53,7 +53,7 @@ Interface file:
     val defaultValue: value: 'T -> result: Result<'T, 'Error> -> 'T
 
     [<CompiledName("DefaultWith")>]
-    val defaultWith: defThunk: (unit -> 'T) -> result: Result<'T, 'Error> -> 'T
+    val defaultWith: defThunk: ('Error -> 'T) -> result: Result<'T, 'Error> -> 'T
 
     [<CompiledName("Count")>]
     val count: result: Result<'T, 'Error> -> int


edgarfgp:
@cartermp would love to have a review please ?

cartermp:
Thanks for the ping. I think this looks fine but would like @dsyme to give it a final look.

vzarytovskii:
I don't remember if the current API was already released to nuget. If it did, this will technically be a breaking change (unit vs generic param).

edgarfgp:
Hope we can fix this before released to nuget. If not we can add another function that adds this ? . I will more that happy to help either way :)

cartermp:
@vzarytovskii I don't think it's in any shipped version, no. So there should be time? The PR that added the feature set was done in June 20th

dsyme:
I'm ok with this as long as it hasn't been shipped to nuget

vzarytovskii:
> I'm ok with this as long as it hasn't been shipped to nuget

It hasn't been released as a nuget package but has been part of SDK/VS since June 20th(ish). Also, we have a beta package released about 2 weeks ago, as long as we fine with breaking betas, it's OK.

cc @KevinRansom

cartermp:
By that you mean preview VS, right? Design changes like this are fine to make in previews I'd say (at least that's certainly the design for language and IDE features).

gusty:
@witoldsz either that (previews) or maybe you are using FsToolkit

witoldsz:
I am pretty sure the version with `Error _ -> defThunk ()` is in official dotnet/F# 6 version.

Let me just check:
```
$ dotnet --version
6.0.417
```
```
$ dpkg -l |grep dotnet-sdk
ii  dotnet-sdk-6.0                              6.0.417-1                                   amd64        Microsoft .NET SDK 6.0.417
```

The issue came up when my colleague tried to upgrade services to work with dotnet 8 and one stopped compiling because of the new meaning of the `Result.withDefault`.

PR #700, opened 2022-08-10
Update RFC FS-1126 Allow lowercase DU cases when RequireQualifiedAccess

edgarfgp:

### DIFF
--- a/RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
+++ b/RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
@@ -4,7 +4,7 @@ The design suggestion [Allow lower-case DU cases when `[<RequireQualifiedAccess>
 
 This RFC covers the detailed proposal for this suggestion.
 
-- [x] [Suggestion]((https://github.com/fsharp/fslang-suggestions/issues/131)
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/131)
 - [x] Approved in principle
 - [x] [Implementation](https://github.com/dotnet/fsharp/pull/13432)
 - [ ] Design Review Meeting(s) with @dsyme and others invitees
@@ -35,18 +35,21 @@ Note: the above example is not a valid type name for .NET, so it will remain as 
 # Detailed design
 This RFC will avoid a compiler check for union case names if the `[<RequireQualifiedAccess>]` is used at the type level.
 
+We will reuse error FS0053 with an updated message : "Lowercase discriminated union cases are only allowed when using RequireQualifiedAccess attribute"
+
+The following examples will now be valid :
+
 Example code:
 
 ```fsharp
 [<RequireQualifiedAccess>]
-type DU = a 
-
 type DU = a
 
 [<RequireQualifiedAccess>]
 type DU = | a
 
-type DU = | a
+[<RequireQualifiedAccess>]
+type DU = ``a``
 
 [<RequireQualifiedAccess>]
 type DU =
@@ -55,18 +58,6 @@ type DU =
      | C
      | ``D`` of bool
      | ``d``
-     
- [<RequireQualifiedAccess>]
- type DU = ``a``
-
- [<RequireQualifiedAccess>]
- type DU = ``A``
-
- [<RequireQualifiedAccess>]
- type DU = | ``a``
-
- [<RequireQualifiedAccess>]
- type DU = | ``A``
 ```
 
 # Drawbacks
--- a/RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
+++ b/RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
@@ -18,16 +18,21 @@ Currently is not allowed to define a lower-case DU. This is to prevent ambiguity
 However, this is not an issue if the `[<RequireQualifiedAccess>]` attribute is specified on the DU.
 
 ```fs
+// type with [<RequireQualifiedAccess>] attribute is not allowed to be lower-case
 [<RequireQualifiedAccess>]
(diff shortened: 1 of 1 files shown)


RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
```diff
@@ -6,7 +6,7 @@ This RFC covers the detailed proposal for this suggestion.
 
 - [x] [Suggestion]((https://github.com/fsharp/fslang-suggestions/issues/131)
```
abelbraaksma, line 7: Typo, there’s an extra opening parentheses here (was already there, but maybe you can fix it as part of this PR).

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
```diff
@@ -28,6 +28,9 @@ type DU = | a // error FS0053: Discriminated union cases and exception labels mu
 
 type DU = | a // error FS0053: Discriminated union cases and exception labels must be uppercase identifiers
```
abelbraaksma, line 29: These two lines are identical and below there’s something similar with one with and one without `RequireQualifiedAccess`, but no explanation of the difference.

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
```diff
@@ -44,6 +47,26 @@ type DU = a
 type DU = | a
 
 type DU = | a
```
abelbraaksma, line 49: We should explain the difference of these seemingly identical line, maybe with a comment that the top one doesn’t, and bottom one will raise an exception during compiling.

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
```diff
@@ -28,6 +28,9 @@ type DU = | a // error FS0053: Discriminated union cases and exception labels mu
 
 type DU = | a // error FS0053: Discriminated union cases and exception labels must be uppercase identifiers
 ```
+type DU = | ``not.allowed`` // error FS0883: Invalid namespace, module, type or union case name
+
+Note: the above example is not a valid type name for .NET, so it will remain as an error.
 
 # Detailed design
 This RFC will avoid a compiler check for union case names if the `[<RequireQualifiedAccess>]` is used at the type level.
```
abelbraaksma, line 36: Can we add a comment on whether or not, and why (if not), there will be a warning available (level 5) that informs users that lower-case DUs are being used? I can imagine that in general people would want to guard against this style of programming. Maybe this was discussed and dismissed, but then a line explaining that would be nice.

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
```diff
@@ -44,6 +47,26 @@ type DU = a
...
+     | B of string
+     | C
+     | ``D`` of bool
+     | ``d``
+     
+ [<RequireQualifiedAccess>]
+ type DU = ``a``
+
+ [<RequireQualifiedAccess>]
+ type DU = ``A``
```
abelbraaksma, line 63: I don’t see the merit of these examples as they’re already valid (the bottom one, I mean). Can you add a comment as to how these relate to this RFC?

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
```diff
@@ -44,6 +47,26 @@ type DU = a
...
+
+ [<RequireQualifiedAccess>]
+ type DU = | ``a``
+
+ [<RequireQualifiedAccess>]
+ type DU = | ``A``
```
abelbraaksma, line 69: These are identical to the ones above. What’s the merit, what’s the difference, if any? A comment may help.

abelbraaksma (changes requested):
Some suggestions and a few typos I found, see below. Great work on the implementation, btw!

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
edgarfgp: I think you are right :) . As this is RFC is about lowercase DU I thought adding an example with Uppercase might help . But based on your comment I guess not :)

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
edgarfgp: I think you are right :) . As this is RFC is about lowercase DU I thought adding an example with Uppercase might help . But based on your comment I guess not :)

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
edgarfgp: This feature is currently in preview and user will see "Lowercase discriminated union cases are only allowed when using RequireQualifiedAccess attribute" . And when it is out of preview user will see the same error message reusing the same FS0053 in the case of when using lowercase without the use  use of RequireQualifiedAccess .

RFCs/FS-1126-allow-lower-case-du-cases when-require-qualified-access-is specified.md
edgarfgp: Fixed

edgarfgp:
@abelbraaksma I updated the PR based on your suggestions :) . Thanks for taking the time :)

edgarfgp:
@dsyme would love to have a review with this additions ? . Thanks in advance

dsyme (approved):

PR #702, opened 2022-08-16
Update FS-1125-print-println-functions.md

albert-du:
Update rfc to reflect discussion in dotnet/fsharp/pull/13597

### DIFF
--- a/RFCs/FS-1125-print-println-functions.md
+++ b/RFCs/FS-1125-print-println-functions.md
@@ -13,30 +13,33 @@ This RFC covers the detailed proposal for this suggestion.
 # Summary
 
 This RFC introduces two new functions:
-- `print : string -> unit`
-- `println : string -> unit`
+- `print<'T> : 'T -> unit`
+- `println<'T> : 'T -> unit`
 
-`print` writes the exact input to stdout (analogous to `System.Console.Write`) and `println` appends a newline to the string before writing to stdout (analogous to `System.Console.WriteLine`).
+`print` writes the exact input to stdout (analogous to `System.Console.Write`) and `println` appends a newline when writing to stdout (analogous to `System.Console.WriteLine`).
 
 # Motivation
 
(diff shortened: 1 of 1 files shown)


RFCs/FS-1125-print-println-functions.md
```diff
@@ -13,30 +13,33 @@ This RFC covers the detailed proposal for this suggestion.
 # Summary
 
 This RFC introduces two new functions:
-- `print : string -> unit`
-- `println : string -> unit`
+- `print<'T> : 'T -> unit`
+- `println<'T> : 'T -> unit`
 
-`print` writes the exact input to stdout (analogous to `System.Console.Write`) and `println` appends a newline to the string before writing to stdout (analogous to `System.Console.WriteLine`).
+`print` writes the exact input to stdout (analogous to `System.Console.Write`) and `println` appends a newline when writing to stdout (analogous to `System.Console.WriteLine`).
```
abelbraaksma, line 19: Now that the input type changes, the phrase `writes the exact input to stdout` becomes a bit dubious. I.e., what does it mean for writing `Some 42`? Or `print None`? What happens when we write `let f x = 42; print f`?

The simplest answer is probably: whatever `Console.Write[Line]` would do. Which means: casting it to `obj` (that's the only matching overload) and printing it.

But if you check my discussion with @KevinRansom, it really depends on your viewpoint. Do you want it to behave similar to `printfn` or more similar to `Console.WriteLine`? My personal preference would be that it behaves like:

```
let println x =
   printfn "%A" x 
   |> Console.WriteLine
```

This would now print `None` for `None` and `Some 42` for `Some 42`. Without this, it would print `""` (nothing) for `None` and `Some(42)` for `Some 42`.

Similar differences exist for other types. And then there's the whole discussion of whether we use `InvariantCulture` like with the `string` function, or we use the current culture. This changes how `DateTime`, `int` and `float` are printed.

Just a suggestion: perhaps we should take what you currently have and allow others to chime in on this first? Your change is good and adds already a lot of value for newcomers.

abelbraaksma (changes requested):
A few comments/questions/suggestions, see below:

albert-du:
following further discussion in dotnet/fsharp#13597, accepting a generic type doesn't seem to be the way to go

abelbraaksma:
Agreed.

dsyme:
> following further discussion in [dotnet/fsharp#13597](https://github.com/dotnet/fsharp/pull/13597), accepting a generic type doesn't seem to be the way to go

It's not that it's wrong - it's just a very major unresolved issue  :))

PR #709, opened 2022-09-20
Create FS-1128-allow-static-members-in-interfaces.md

gusty:

### DIFF
--- a/RFCs/FS-1128-allow-static-members-in-interfaces.md
+++ b/RFCs/FS-1128-allow-static-members-in-interfaces.md
@@ -1,0 +1,49 @@
+# F# RFC FS-1128 - Allow static members in interfaces
+
+The design suggestion [Allow static members in interfaces](https://github.com/fsharp/fslang-suggestions/issues/1191) has been marked "approved in principle".
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/1191)
+- [x] Approved in principle
+- [ ] Implementation
+- [ ] Design Review Meeting(s) with @dsyme and others invitees ???
+- [ ] Discussion ???
+
+# Summary
+Allow static members in interfaces
+
+# Motivation
+Currently is not allowed to define a static member within interfaces, however the CLR supports this and C# as well since C# 8.
+
+F# can consume those static members and SRTPs can "see" them (as opposed to extension members).
+
+```fs
+
+[<Interface>]
+type IAdditionOperator<'T> =
+    abstract myMember: 'T * 'T -> 'T
+    static member myStaticMember x = x  // error FS0868: Interfaces cannot contain definitions of concrete members. You may need to define a constructor on your type to indicate that the type is a class.
+
+```
+
+# Detailed design
+This RFC will avoid a compiler check for static members in interfaces and make sure the member is properly compiled.
+
+
+# Drawbacks
+None
+
+# Alternatives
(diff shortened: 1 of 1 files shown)


RFCs/FS-1128-allow-static-members-in-interfaces.md
```diff
@@ -0,0 +1,49 @@
...
+
+The design suggestion [Allow static members in interfaces](https://github.com/fsharp/fslang-suggestions/issues/1191) has been marked "approved in principle".
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/1191)
+- [x] Approved in principle
+- [ ] Implementation
+- [ ] Design Review Meeting(s) with @dsyme and others invitees ???
+- [ ] Discussion ???
```
abelbraaksma, line 11: Typically, you'd create a discussion thread (click "Discussion") and add the link here.

RFCs/FS-1128-allow-static-members-in-interfaces.md
```diff
@@ -0,0 +1,49 @@
...
+
+[<Interface>]
+type IAdditionOperator<'T> =
+    abstract myMember: 'T * 'T -> 'T
+    static member myStaticMember x = x  // error FS0868: Interfaces cannot contain definitions of concrete members. You may need to define a constructor on your type to indicate that the type is a class.
+
+```
+
+# Detailed design
+This RFC will avoid a compiler check for static members in interfaces and make sure the member is properly compiled.
```
abelbraaksma, line 31: I think we need to define "properly compiled" here. Do we follow the C# approach? Can we pick details from their approach?

I also think we need to answer a bunch of questions (from the top of my head, not sure if everything needs to be answered):

* What happens when we inherit a static interface?
* How will `default` interop here?
* Are static properties allowed and how do we define `get|set`
* Are static indexers allowed?
* Can we combine with `Abstract` or does it have to be an `Interface`
* You use `[<Interface>]`, I assume it also works with `type TT = [...] interface end`? 
* How is implementing an interface with static members going to work?
* How do we deal with the standard of F# for explicit interfaces, is that relevant here?
* How is the diamond problem dealt with (it's different from normal interfaces, as we don't have an instance)?
* From all C# features on static interfaces, what parts are we going to support and what parts aren't we?
* Are all members public?
* How does the IL-compiled structure look like? (I assume we can link to C# here, there must be a design somewhere)

abelbraaksma (changes requested):
A couple of questions, though some of them could be just logged, and answered later, I guess. I also think we need more concrete examples here, both from defining (multiple/single inheritance?) and consuming (how does the implementing syntax look?).

RFCs/FS-1128-allow-static-members-in-interfaces.md
dsyme: Could you add all these questions and resolutions to the RFC please?

dsyme: I think we could then merge this RFC

gusty: Done !

dsyme (approved):

dsyme:
@gusty Thanks for this!   Note also the similarity with https://github.com/dotnet/fsharp/pull/14132, the RFC for that is TBD

gusty:
I see, maybe both RFCs can be implemented in that PR.

PR #710, opened 2022-09-21
Create FS-1129-shorthand-anonymous-unary-functions

tboby:
This is just a starting point for https://github.com/fsharp/fslang-suggestions/issues/506 with a basic proof of concept implementation found at https://github.com/dotnet/fsharp/pull/13907

### DIFF
--- a/RFCs/FS-1129-shorthand-anonymous-unary-functions.md
+++ b/RFCs/FS-1129-shorthand-anonymous-unary-functions.md
@@ -1,0 +1,154 @@
+# F# RFC FS-1129 - (`_.` shorthand for simple anonymous unary functions)
+
+The design suggestion [Allow _.Property shorthand for accessor functions](https://github.com/fsharp/fslang-suggestions/issues/506) has been marked "approved in principle".
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/506)
+- [x] Approved in principle
+- [ ] [Implementation](https://github.com/dotnet/fsharp/pull/13907)
+- [ ] Design Review Meeting(s) with @dsyme and others invitees
+- [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)
+
+# Summary
+
+Add a syntactic sugar shorthand (`_.`) for simple functions consisting of accessors and atomic function application only.
+
+```fsharp
+_.Foo  => fun o -> o.Foo
+```
+
+# Motivation
+
+It is fairly common to need to create function expressions which consist of a single argument and only "dot" operations upon it.
+
+```fsharp
+someList |> List.map (fun x -> x.ToString())
+someList |> List.exists (fun x -> x.Contains "foo")
+someList |> List.sortBy (fun x -> x.Bar)
+```
+
+These functions:
+
+- Add a small amount of visual noise
+- Add a small amount of cognitive overhead and maintenance burden on naming the function argument
+- Have a small risk of leading to errors if the wrong ident is used on the rhs
+
+The introduction of a dedicated syntax for these functions could improve the general experience when piping values.
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Detailed design
+
+This is a syntactical transformation of the form:
+
+```fsharp
+_.Foo  => fun o -> o.Foo
+```
+
+And the following things should be valid:
+
+```fsharp
+_.Foo.Bar  => fun o -> o.Foo.Bar
+_.Foo.[5]  => fun o -> o.Foo.[5]
+_.Foo()    => fun o -> o.Foo()
+_.Foo(5).X => fun o -> o.Foo(5).X
+```
+
+This transformation should take place after the SyntaxTree is created, during the checking of the expression.
+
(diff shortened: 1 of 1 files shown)


RFCs/FS-1129-shorthand-anonymous-unary-functions.md
```diff
@@ -0,0 +1,208 @@
...
+
+```fsharp
+_.Foo  => fun o -> o.Foo
+```
+
+And the following things should be valid:
+
+```fsharp
+_.Foo.Bar  => fun o -> o.Foo.Bar
+_.Foo.[5]  => fun o -> o.Foo.[5]
```
dsyme, line 53: Please update for new indexer syntax

RFCs/FS-1129-shorthand-anonymous-unary-functions.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Culture-aware formatting/parsing
+
+TODO, but no
+
+Does the proposed RFC interact with culture-aware formatting and parsing of numbers, dates and currencies? For example, if the RFC includes plaintext outputs, are these outputs specified to be culture-invariant or current-culture.
+
+# Unresolved questions
+
+1. Is underscore definitely the appropriate symbol?
```
dsyme, line 202: I regard (1) as resolved unless there are specific problems encountered

RFCs/FS-1129-shorthand-anonymous-unary-functions.md
```diff
@@ -0,0 +1,208 @@
...
+2. Is there a good name for this feature?
+3. What happens when you nest this?
+4. Are you allowed inline comments between `_` and `.` or between `.` and `expr`?
```
dsyme, line 205: This sort of thing is important to clarify

RFCs/FS-1129-shorthand-anonymous-unary-functions.md
```diff
@@ -0,0 +1,208 @@
...
+
+TODO, but no
+
+Does the proposed RFC interact with culture-aware formatting and parsing of numbers, dates and currencies? For example, if the RFC includes plaintext outputs, are these outputs specified to be culture-invariant or current-culture.
+
+# Unresolved questions
+
+1. Is underscore definitely the appropriate symbol?
+2. Is there a good name for this feature?
+3. What happens when you nest this?
```
dsyme, line 204: At the moment, as specified, no significant confusing with regard to nesting appears possible.

kerams:
Is intra-assembly closure reuse feasible?

```fsharp
let r1: Record = ...
let r2: Record = ...
let r1x = r1 |> Option.map _.x
let r2x = r2 |> Option.map _.x
```

There should ideally be only one generated closure per property on type in an assembly.

vzarytovskii:
> Is intra-assembly closure reuse feasible?
> 
> ```fsharp
> let r1: Record = ...
> let r2: Record = ...
> let r1x = r1 |> Option.map _.x
> let r2x = r2 |> Option.map _.x
> ```
> 
> There should ideally be only one generated closure per property on type in an assembly, and it could be given a descriptive name. 

Behaviour should be no different to how full lambda works now, since we just rewrite _.Prop to regular lambda in untyped ast.

T-Gro:
But even the full lambda as of now produces duplicates for identical operations.
https://sharplab.io/#v2:DYLgZgzgPg9gDgUwHYAIDKBPCAXBBbAWAChjjgFsUAnARgA9qaUBeRlKAPhQHk5sBLGEgB0eAIZwUACjABXVAwC0XOsIAqMNNir8kAcykBKQ2QrUATAyrmWF9l14ChoidLkKUylKo1ad+o0MgA==

I believe the comment from @kerams  is valid && orthogonal => might be worth a separate suggestions for the optimizer to de-duplicate identical lambda-generated classes.

vzarytovskii:
> But even the full lambda as of now produces duplicates for identical operations.
> https://sharplab.io/#v2:DYLgZgzgPg9gDgUwHYAIDKBPCAXBBbAWAChjjgFsUAnARgA9qaUBeRlKAPhQHk5sBLGEgB0eAIZwUACjABXVAwC0XOsIAqMNNir8kAcykBKQ2QrUATAyrmWF9l14ChoidLkKUylKo1ad+o0MgA==
> 
> I believe the comment from @kerams  is valid && orthogonal => might be worth a separate suggestions for the optimizer to de-duplicate identical lambda-generated classes.

I honestly think we shouldn't be doing such analysis, since it can be quite complex with little benefits.

smoothdeveloper:
The feature is going to be shipping but it is very difficult to get access to the RFC as it wasn't merged and the suggestion description doesn't mention it. Also no discussion thread was opened for the RFC itself.

I'm mostly concerned seeing the feature implementation is not restricted to properties only (which was Don main restraint for initial take on the feature) and will have a few rough edges on release in terms of tooling (which the F# users should be ok with, as things will be sorted out); but overall, we should make sure the suggestion discussion is closed in favor to a discussion related to the RFC, and that the RFC (even if not the last one) is visible in the repository itself, with the link to the discussion thread.

I'd prefer an almost empty RFC in the repository, with link to the discussion thread, and the suggestion being closed (with top post updated with links, as well as the closing comment), than the current situation.

PR #715, opened 2022-10-25
RFC for triple quoted string interpolation

abonie:
[Rendered RFC](https://github.com/abonie/fslang-design/blob/fs-1128-triple-quoted-interpolation/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md)

### DIFF
--- a/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
+++ b/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
@@ -43,15 +43,19 @@ This change should:
 The existing form of interpolated triple quoted strings: `$"""...{}..."""` is extended to allow multiple dollar signs at the begining of the string literal.
 The count of these initiatory `$` characters indicates how many `{` and `}` characters are used to delimit interpolation expression within the content of the literal.
 
-Important note: To maintain backward compatibility, there are no changes in semantics for a single-`$` case - in particular, curly braces can still be escaped by doubling them.
-This means that it is an edge case in terms of how this feature behaves.
-It would be nice to have a clean design that covers all cases, but it is not worth breaking backward compatibility for.
-Therefore the below description applies to the literals starting with 2 or more `$`.
+Behavior of triple quoted string literals starting with a single `$` remains unchanged.
 
-Triple quoted string literals starting with 2 or more `$` characters have no escaping mechanism for `{` or `}`.
-Instead, the literal can always be constructed in such a way as to ensure that interpolation delimiters will not collide with other curly braces in the content.
-In a literal that starts with `N` `$` characters, a sequence of consecutive `{` can be at most `2*N-1` characters long and for any sequence of more than `N`, the innermost `N` `{` characters delimit the interpolation, while the outter up to `N-1` are treated as content (analogously for `}`).
-A sequence of `2*N` or more `{` (or `}`) will result in a compilation error.
+Triple quoted string literals starting with `N` `$` (where `N` > 1) behave as follows:
+- No escaping mechanism for any characters.
+- A sequence of `N` `{` indicates the beginning of an interpolation expression and a sequence of `N` `}` indicates the end of interpolation expression.
+- A sequence of `2*N` or more `{` or `}` is not allowed and will result in a compilation error.
+- In any sequence of more than `N` `{`, the outter braces are treated as content of the string, while the innermost `N` delimit the interpolation (and analogously for `}`).
+
+Note: This design is taken from C#'s [raw string literals](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md).
+It is an elegant solution that will already be familiar to a portion of dotnet developers.
+However, there is one significant difference - in raw strings, these rules also extend to literals with single `$`.
+Unfortunately, in F# that would be a breaking change,
+so single `$` case remains an exception in terms of alignment with C#.
 
 Example:
 
@@ -73,8 +77,9 @@ $$$"""{{{{41+1}}}} = {42}"""
 # Drawbacks
 
 Even though this only extends already existing syntax, it still can be considered yet another way of doing string interpolation.
-Moreover, it can't be fully aligned with analogous C# feature (raw string literals - [docs](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals)) and stay backward compatible at the same time, so this could be a source of confusion
-(but note that triple quoted string literals are already similar but not really aligned with C#'s raw string literals anyway - see [spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md)).
+
+Moreover, as mentioned in [Detailed design section](#detailed-design) it can't be fully aligned with analogous C# feature (raw string literals - [docs](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals)) and stay backward compatible at the same time, so this could be a source of confusion
+(note that triple quoted string literals are already not fully aligned with C#'s raw string literals anyway - see [detailed spec for raw strings](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md)).
 
 # Alternatives
 
@@ -114,7 +119,7 @@ This feature should have no impact on performance.
 
 ## Scaling
 
-We do not set an explicit limit on the number of `$` signs allowed at the beginning of a string literal.
+We do not set an explicit limit on the number of `$` characters allowed at the beginning of a string literal.
 In a hand-written code it will typically not exceed 3.
 
 ## Culture-aware formatting/parsing
--- a/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
+++ b/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
@@ -81,7 +81,43 @@ $$$"""{{{{{41+1}}}} = {{42}"""
 Even though this only extends already existing syntax, it still can be considered yet another way of doing string interpolation.
 
 Moreover, as mentioned in [Detailed design section](#detailed-design) it can't be fully aligned with analogous C# feature (raw string literals - [docs](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals)) and stay backward compatible at the same time, so this could be a source of confusion
-(note that triple quoted string literals are already not fully aligned with C#'s raw string literals anyway - see [detailed spec for raw strings](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md)).
+(note that triple quoted string literals are already not fully aligned with C#'s raw string literals anyway - see [next section](#comparison-with-raw-strings).
+
+# Comparison with raw strings
+
+Raw string literals is a feature in C# corresponding to this proposal see - [detailed spec for raw strings](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md)).
+
+For convenience, here is a list of features that raw strings support and that are missing/different from triple quoted strings:
+- Multi-line raw string literals in C# must start with a new line. They ignore the first and last new line, so the below literal will be 2 lines in C#, but 4 lines (including two "empty" lines) in F#:
+```
+"""
+This is the first line in C#, but second line in F#.
+This is the second and last line in C#, but third and second-to-last line in F#.
+"""
+```
+- Raw strings in C# de-indent its content based on the indentation of the last line. This is to make literals in code more readable. For example:
+```
+var xml = """
+          <element attr="content">
+            <body>
+            </body>
+          </element>
+          """;
+```
+Since last line in the literal is indented by 10 spaces, each line in the contents gets de-indented by that much, resulting in:
+```
+<element attr="content">
+  <body>
+  </body>
+</element>
+```
+- Raw string in C# can start with more than three `"` characters. It allows for sequences of more `"` in the contents of the string (similarly to `$` and curly braces):
+```
+""""
+This literal starts with 4x" so it can have """ in its contents.
+It has to end with 4x" to match the opening.
+""""
+```
 
 # Alternatives
 
--- a/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
+++ b/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
@@ -8,7 +8,7 @@ This RFC covers the detailed proposal for this suggestion.
 - [x] Approved in principle
 - [ ] [Implementation]() (no implementation yet, but a proof of concept [here](https://github.com/dotnet/fsharp/compare/main...abonie:fsharp:poc_improved_interpolation))
 - [ ] Design Review Meeting(s) with @dsyme and others invitees
-- [Discussion]()
+- [Discussion](https://github.com/fsharp/fslang-design/discussions/716)
 
 # Summary
 
--- a/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
+++ b/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
@@ -129,6 +129,7 @@ N/A
 # Unresolved questions
 
 1. Do we want to also provide a way to easily input percent signs in interpolated string literals?
+2. Should we extend this feature to regular (single-quoted) string literals as well in addition to triple quoted strings?
 
 # Links
 
--- a/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
+++ b/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
@@ -86,8 +86,12 @@ Moreover, as mentioned in [Detailed design section](#detailed-design) it can't b
 Alternatives:
 
 1. Try to fully align with C#'s raw string literals at the cost of backward compatiblity.
+
+  Such change would break any code that has at least one escaped `{` or `}` in an interpolated (triple quoted) string literal. This feature does not warrant such a breaking change.
+
 2. Introduce alternative syntax for delimiting expression fills that does not clash with `{` and `}` which are common in JSON and certain other contexts.
-Discussed [here](https://github.com/fsharp/fslang-design/discussions/716#discussioncomment-4039580).
+
+  Discussed [here](https://github.com/fsharp/fslang-design/discussions/716#discussioncomment-4039580).
 
 # Compatibility
 
--- a/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
+++ b/RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
@@ -13,6 +13,7 @@ This RFC covers the detailed proposal for this suggestion.
 # Summary
 
 Extend current form of interpolated triple quoted string to allow specifying how many braces should start an interpolation (equal to the number of dollar signs at the start of the literal). For example:
+
 ```
 let json = $$"""
 {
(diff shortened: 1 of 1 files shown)


RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+# Compatibility
+
+Please address all necessary compatibility questions:
+
+* Is this a breaking change?
+*No*
+* What happens when previous versions of the F# compiler encounter this design addition as source code?
+*A compiler error*
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+*AFAICT it should work, changes will hopefully be limited to lexing stage*
```
cartermp, line 74: Yes, this would likely be compatible. Or at least I see no reason why it wouldn't.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+Important note: To maintain backward compatibility, there are no changes in semantics for a single-`$` case. This means that it is an edge case in terms of how this feature behaves. It would be nice to have a clean design that covers all cases, but it is not worth breaking backward compatibility for. Therefore the below description applies to the literals starting with 2 or more `$`.
+
+Triple quoted string literals starting with 2 or more `$` characters have no escaping mechanism for `{` or `}`. Instead, the literal can always be constructed in such a way as to ensure that interpolation delimiters will not collide with other curly braces in the content. In a literal that starts with `N` `$` characters, a sequence of consecutive `{` can be at most `2*N-1` characters long and for any sequence of more than `N`, the innermost `N` `{` characters delimit the interpolation, while the outter up to `N-1` are treated as content (analogously for `}`). A sequence of `2*N` or more `{` (or `}`) will result in a compilation error.
```
abelbraaksma, line 42: Can you break these long single line paragraph sentences up by adding line breaks? It’ll remain rendered as a single oars, but will help with adding targeted comments, and is easier on the diffs and people editing or viewing the source.

I needed to read this sentence three times. Though I don’t immediately have a clear simpler version in mind. Maybe this came from the C# spec?

Perhaps we could add a bunch of smaller examples with more of the variants shown and how they relate to 2N-1.

Also, is there a limit? I.e., what does the compiler do when encountering, say, 100 leading $ signs?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+Example:
+```
+let text = "3 curly braces delimit interpolation expression"
+$$$"""
+In a string literal with 3x$, { or {{ are simply treated as content
+whereas {{{text}}}.
+It is also allowed to have {, {{, } or }} adjacent to the interpolation delimiters as so:
+{{{{41+1}}}} = {42}
```
abelbraaksma, line 51: Suggestion: I’d prefer an example in the style of FSI prompts, so we can more easily follow along and see what the code does?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+This change should:
+1. be backwards compatible with existing support for triple quoted strings
+2. align with C# raw string literals reasonably well
+
+# Detailed design
+
+The existing form of interpolated triple quoted strings: `$"""...{}..."""` is extended to allow multiple dollar signs at the begining of the string literal. The count of these initiatory `$` characters indicates how many `{` and `}` characters are used to delimit interpolation expression within the content of the literal.
+
+Important note: To maintain backward compatibility, there are no changes in semantics for a single-`$` case. This means that it is an edge case in terms of how this feature behaves. It would be nice to have a clean design that covers all cases, but it is not worth breaking backward compatibility for. Therefore the below description applies to the literals starting with 2 or more `$`.
```
abelbraaksma, line 40: What behaviour are you considering an edge case? Edge case compared to C# perhaps, and how does it differ? (I understand it stays the same, so that sounds like there is no edge case, right?)

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+Compiler tooling should offer the same support for the extended form of interpolated string literals as it currently does for interpolated triple quoted literals. This includes colorization, autocompletion, and navigational features in an interpolated context.
+
+## Performance
+
+AFAICT there should be no impact on performance (although there might be if using certain constructs in regexes within .fsl files can have a big impact `XXX TODO`)
+
+## Scaling
+
+Please list the dimensions that describe the inputs for this new feature, e.g. "number of widgets" etc.  For each, estimate a reasonable upper bound for the expected size in human-written code and machine-generated code that the compiler will accept.
```
abelbraaksma, line 95: I don’t understand “number of widgets” in this context. Also, why does this start with “please list…”? Is this a left over from the template?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Culture-aware formatting/parsing
+
+Does the proposed RFC interact with culture-aware formatting and parsing of numbers, dates and currencies? For example, if the RFC includes plaintext outputs, are these outputs specified to be culture-invariant or current-culture.
+
+*No* `TODO: remove this section?`
+
+# Unresolved questions
+
+1. Do we want to also provide a way to easily input procent signs in interpolated string literals?
```
abelbraaksma, line 114: I think it’s “percent”, not “procent”? Though I don’t have an answer to your question ;). Seems nontrivial to support.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+In a string literal with 3x$, { or {{ are simply treated as content
+whereas {{{text}}}.
+It is also allowed to have {, {{, } or }} adjacent to the interpolation delimiters as so:
+{{{{41+1}}}} = {42}
+"""
+```
+
+# Drawbacks
+
+Even though this only extends already existing syntax, it still can be considered yet another way of doing string interpolation. Moreover, it can't be fully aligned with analogous C# feature (raw string literals) and stay backward compatible at the same time, so this could be a source of confusion (but note that triple quoted string literals are already similar but not really aligned with C#'s raw string literals anyway).
```
abelbraaksma, line 57: You mention “raw string literals” several times. Can we briefly explain them? Assume us fsharpers don’t know that feature (I don’t). How are they not aligned?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Alternatives
+
+Alternatives:
+1. Try to fully align with C#'s raw string literals at the cost of backward compatiblity.
```
abelbraaksma, line 62: We should probably say why were not doing these alternatives. If there’s existing discussion and consensus we can link to the discussion.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Motivation
+
+Triple quoted strings are already quite handy for embedding e.g. XML in string literals, but not as convenient for embedding some other languages that use `{` and `}` in their syntax, e.g. JSON or CSS. Currently every occurrence of a curly brace in an interpolated string needs to be escaped, which can quickly become tedious. With this extended syntax, only the braces around interpolated expression need special attention, and that is where programmer's focus already is.
+
+The motivation is not to enable any novel use cases, but to make working with interpolated string literals a much more seamless experience in the use cases mentioned above.
+
+This change should:
+1. be backwards compatible with existing support for triple quoted strings
+2. align with C# raw string literals reasonably well
```
abelbraaksma, line 34: Can we link and/or explain this term?

abelbraaksma (changes requested):
Some suggestions below. Most of them just requests for clarifications in the text and a bit of formatting.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: Yep, this part is left from a template, I only added a line at the bottom of this section - still need to figure this out

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: Hmm... You are right, this is not very clear.

If you think only about the F# part (forget about C# for now), trying to explain how adding more `$` at the beginning of the literal affects parsing the contents with respect to `{` and `}`, you could pretty much come up with an inductive rule, but it would have to have double `$` as a base case, and single `$` would be a special case (because it has a way of escaping `{` and `}` and has no limit to how many are allowed in a sequence)

In C# in contrast, there is no special case for single `$`, it follows exactly the same rules as multiple `$`: no escaping mechanism for curly braces, up to `2*N-1` consecutive `{` (or `}`) allowed.

I will think about how to rewrite this part to make it clearer.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: There are links at the bottom of this RFC with description of raw string literals, I can refer to them here.

If you think there should be some direct explanation of raw string literals in this RFC and how they differ from triple quoted strings, I can think about that, but I'd argue it shouldn't go in this particular section anyway.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: There's not, should I start a discussion using the GitHub discussions feature, or keep it in the [suggestion issue](https://github.com/fsharp/fslang-suggestions/issues/1150)?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
T-Gro: In C#, even 111 of dollar signs still works:

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoANwEMoACAWwE8A1BgSzuABsYAOTrMYjALyMArAG5G1ekzYBlDFC54A5hMYASfQcNHjJ02fMXLV68YBE9gCoALLgGdGblq0YYYCDADeQcEhoWHhEZFR0TGxcWFsnOq8AsKiAL6ZWdk5uXn5BYVFxSWlufa2MtTURGQAnAAUKmoamgCUMkA===

abonie: There's no hardcoded limit, and I don't think there's a limit mentioned in C# documentation. Maxint amount of dollar signs would definitely be problematic. I think Scaling section is meant for such details, but I haven't filled it out yet.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: I am afraid you are right about it being nontrivial. At least it is not a natural extension of changes that are needed for the main part of this RFC and it seems like it wouldn't be limited to lexing either.

dsyme:
This is looking good!

@abelbraaksma Can I ask you to be primary reviewer on this? When we get a green approve from you I can take a final look.  But basically I'm good with the general direction of this, and love how you're helping refine the details.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: I have split it into single line strings, because it started looking awkward with multiline literal + output. Added FSI style output as `//` comments. Let me know if it looks good now or maybe I should come up with different examples?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abelbraaksma: Typically a discussion thread would be opened for rfcs

abelbraaksma:
@dsyme, sure thing, happy to help!

abonie:
Hey @abelbraaksma, thanks for the review so far. I have tried to rewrite some sections to make them easier to read and applied your other suggestions too, so whenever you have time for round two, feel free :)

I haven't covered why we are not doing any of the alternatives, but I hope to have some discussion about that first - perhaps more exploration is needed.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -21,99 +22,117 @@ let json = $$"""
...
+Currently every occurrence of a curly brace in an interpolated string needs to be escaped, which can quickly become tedious.
+With this extended syntax, only the braces around interpolated expression need special attention, and that is where programmer's focus already is.
 
 The motivation is not to enable any novel use cases, but to make working with interpolated string literals a much more seamless experience in the use cases mentioned above.
 
 This change should:
+
 1. be backwards compatible with existing support for triple quoted strings
-2. align with C# raw string literals reasonably well
+2. align with C# [raw string literals](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md) reasonably well
```
abelbraaksma, line 39: Thanks for this, I have some reading to do :)

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -21,99 +22,117 @@ let json = $$"""
...
 
 # Detailed design
 
-The existing form of interpolated triple quoted strings: `$"""...{}..."""` is extended to allow multiple dollar signs at the begining of the string literal. The count of these initiatory `$` characters indicates how many `{` and `}` characters are used to delimit interpolation expression within the content of the literal.
+The existing form of interpolated triple quoted strings: `$"""...{}..."""` is extended to allow multiple dollar signs at the begining of the string literal.
+The count of these initiatory `$` characters indicates how many `{` and `}` characters are used to delimit interpolation expression within the content of the literal.
```
abelbraaksma, line 44: "initiatory"? Looks like it's a word. I didn't know it :).

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -21,99 +22,117 @@ let json = $$"""
...
+- If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct?
 *It is not a change/extension to FSharp.Core*
 
 # Pragmatics
 
 ## Diagnostics
 
-* New compiler error for interpolated string literals with too many consecutive `{` or `}` characters.
-* Otherwise, same compiler errors as for regular (single-`$`) interpolated strings.
+- New compiler error for interpolated string literals with too many consecutive `{` or `}` characters.
```
abelbraaksma, line 108: This reminds me, we should probably add a line or two above in the description about those `N * 2 - 1` rules. The question I have about it, does it have to be balanced? If it does have to be, we should have a specific error for that. If it doesn't, we should say so:

```f#
let x = $$$"""is this {{{{{ 42 + 1 }}} legal?"""
```

Does this produce the string `"is this {{42 legal"`? Or does it raise an error?

abelbraaksma: Also, can we already take a stab at what the new error would look like (error number tbd)?

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -21,99 +22,117 @@ let json = $$"""
...
-Does the proposed RFC interact with culture-aware formatting and parsing of numbers, dates and currencies? For example, if the RFC includes plaintext outputs, are these outputs specified to be culture-invariant or current-culture.
-
-*No* `TODO: remove this section?`
+N/A
 
 # Unresolved questions
 
-1. Do we want to also provide a way to easily input procent signs in interpolated string literals?
+1. Do we want to also provide a way to easily input percent signs in interpolated string literals?
+2. Should we extend this feature to regular (single-quoted) string literals as well in addition to triple quoted strings?
```
abelbraaksma, line 132: For convenience, and also by the time we reach a consensus, can you add a link to the specific comment in the discussion thread that asks this question or (quite definitively) answers it?

abelbraaksma (changes requested):
@abonie All in all looks very good! Great work! I'll have another final read-through on the rendered file soon. Found a few nits, but nothing major, see below.

Also, please mark previous and future review remarks as "resolved" once you're done with them, preferably with a little note like "done" or "not relevant anymore" or "see discussion there and there" etc, it helps me and others to understand you've tackled them.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -21,99 +22,117 @@ let json = $$"""
...
+
+$$$"""In a string literal with 3x$, { or {{ are simply treated as content..."""
+// val it: string = "In a string literal with 3x$, { or {{ are simply treated as content..."
+
+$$$"""...whereas {{{text}}}."""
+// val it: string = "...whereas 3 curly braces delimit interpolation expression."
+
+// It is also allowed to have {, {{, } or }} adjacent to the interpolation delimiters like so:
+$$$"""{{{{41+1}}}} = {42}"""
+// val it: string = "{42} = {42}"
```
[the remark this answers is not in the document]
abonie: Well, I think this is just the downside of this approach: if the string you are working with already has multiple consecutive curly braces right next to where interpolation expr is going to be, it results in this mess.

That is a case where [this](https://github.com/fsharp/fslang-design/discussions/716#discussioncomment-4039580) alternative approach would be way nicer... However, I don't think that this is a __common__ use case (hopefully).

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: Tbh now I see that this sentence would preserve its meaning and be just as clear without it, so I can just remove the suspicious word

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: No, the extra braces are just string content, so they don't have to be balanced. I'll add a note.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: Done

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: Added an explanation and a link to discussion.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -21,99 +22,117 @@ let json = $$"""
...
 """
 ```
 
 # Drawbacks
 
-Even though this only extends already existing syntax, it still can be considered yet another way of doing string interpolation. Moreover, it can't be fully aligned with analogous C# feature (raw string literals) and stay backward compatible at the same time, so this could be a source of confusion (but note that triple quoted string literals are already similar but not really aligned with C#'s raw string literals anyway).
+Even though this only extends already existing syntax, it still can be considered yet another way of doing string interpolation.
+
+Moreover, as mentioned in [Detailed design section](#detailed-design) it can't be fully aligned with analogous C# feature (raw string literals - [docs](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals)) and stay backward compatible at the same time, so this could be a source of confusion
+(note that triple quoted string literals are already not fully aligned with C#'s raw string literals anyway - see [detailed spec for raw strings](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/raw-string-literal.md)).
```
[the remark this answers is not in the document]
abonie: Probably, I'll see what I can do.

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abonie: Added new error example and a note about (not) balancing extra braces

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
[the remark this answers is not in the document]
abonie: I have added a section, listing main differences. I tried to keep it brief (or I would end up pasting C# spec).

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
abelbraaksma: Perfect, thanks!

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+# Drawbacks
+
+Even though this only extends already existing syntax, it still can be considered yet another way of doing string interpolation.
+
+Moreover, as mentioned in [Detailed design section](#detailed-design) it can't be fully aligned with analogous C# feature (raw string literals - [docs](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals)) and stay backward compatible at the same time, so this could be a source of confusion
+(note that triple quoted string literals are already not fully aligned with C#'s raw string literals anyway - see [next section](#comparison-with-raw-strings).
+
+# Comparison with raw strings
```
abelbraaksma, line 86: Great to have this section!

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,193 @@
...
+
+For convenience, here is a list of features that raw strings support and that are missing/different from triple quoted strings:
+- Multi-line raw string literals in C# must start with a new line. They ignore the first and last new line, so the below literal will be 2 lines in C#, but 4 lines (including two "empty" lines) in F#:
+```
+"""
+This is the first line in C#, but second line in F#.
+This is the second and last line in C#, but third and second-to-last line in F#.
+"""
+```
+- Raw strings in C# de-indent its content based on the indentation of the last line. This is to make literals in code more readable. For example:
```
abelbraaksma, line 98: nit, grammar:

```suggestion
- Raw strings in C# de-indent their content based on the indentation of the last line. This is done to make literals in code more readable. For example:
```

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,193 @@
...
+          """;
+```
+Since last line in the literal is indented by 10 spaces, each line in the contents gets de-indented by that much, resulting in:
+```
+<element attr="content">
+  <body>
+  </body>
+</element>
+```
+- Raw string in C# can start with more than three `"` characters. It allows for sequences of more `"` in the contents of the string (similarly to `$` and curly braces):
```
abelbraaksma, line 114: nit:
```suggestion
- Raw strings in C# can start with more than three `"` characters. It allows for sequences of more `"` in the contents of the string (similarly to `$` and curly braces):
```

RFCs/FS-1128-better-interpolated-triple-quoted-strings.md
```diff
@@ -0,0 +1,193 @@
...
+          </element>
+          """;
+```
+Since last line in the literal is indented by 10 spaces, each line in the contents gets de-indented by that much, resulting in:
+```
+<element attr="content">
+  <body>
+  </body>
+</element>
+```
```
abelbraaksma, line 113: Maybe add a line emphasizing that this is not the case in F#:

````suggestion
```

In F#, the indentation remains in place, that is, the compiler will not remove it.

````

abelbraaksma (changes requested):
Looks good! Just a few very minor nits, otherwise good to go:

abelbraaksma (approved):
Looks good now!

abelbraaksma:
@dsyme, this is ready, can review and/or merge?

dsyme (approved):

dsyme:
Really great work, thank you!

PR #722, opened 2022-11-22
Rename string interpolation RFC

abonie:
There were two FS-1128-... RFCs. Fixed by renaming the more recent one to FS-1132-...


kerams:
1131 is also used:)

PR #732, opened 2023-02-23
FS-1135 - Random functions for collections

Lanayx:
Click “Files changed” → “⋯” → “View file” for the rendered RFC (i.e.: [here](https://github.com/fsharp/fslang-design/blob/ca1e17507c29ea922a86a30f845b87d494a2e050/RFCs/FS-1135-random-functions-for-collections.md)).

Discussion: [in this thread](https://github.com/fsharp/fslang-design/discussions/731)

### DIFF
--- a/RFCs/FS-1135-random-functions-for-collections.md
+++ b/RFCs/FS-1135-random-functions-for-collections.md
@@ -38,23 +38,23 @@ Two functions should be added to each module.
 
 ```fsharp
 // Array module
-val shuffle: array:'T[] -> 'T[]
-val shuffleWith: random:Random -> array:'T[] -> 'T[]
-val shuffleInPlace: array:'T[] -> 'T[]
-val shuffleInPlaceWith: random:Random -> array:'T[] -> 'T[]
+val randomShuffle: array:'T[] -> 'T[]
+val randomShuffleWith: random:Random -> array:'T[] -> 'T[]
+val randomShuffleInPlace: array:'T[] -> 'T[]
+val randomShuffleInPlaceWith: random:Random -> array:'T[] -> 'T[]
 // List module
-val shuffle: list:'T list -> 'T list
-val shuffleWith: random:Random -> list:'T list -> 'T list
+val randomShuffle: list:'T list -> 'T list
+val randomShuffleWith: random:Random -> list:'T list -> 'T list
 // Seq module
-val shuffle: source:'T seq -> 'T seq
-val shuffleWith: random:Random -> source:'T seq -> 'T seq
+val randomShuffle: source:'T seq -> 'T seq
+val randomShuffleWith: random:Random -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.shuffle // [ "Charlie"; "Dave"; "Alice"; "Bob" ]
+let round1Order = allPlayers |> List.randomShuffle // [ "Charlie"; "Dave"; "Alice"; "Bob" ]
 ```
 
 ### Choice
@@ -65,14 +65,14 @@ Two functions should be added to each module.
 
 ```fsharp
 // Array module
-val choice: array:'T[] -> 'T
-val choiceWith: random:Random -> array:'T[] -> 'T
+val randomChoice: array:'T[] -> 'T
+val randomChoiceWith: random:Random -> array:'T[] -> 'T
 // List module
-val choice: list:'T list -> 'T
-val choiceWith: random:Random -> list:'T list -> 'T
+val randomChoice: list:'T list -> 'T
+val randomChoiceWith: random:Random -> list:'T list -> 'T
 // Seq module
-val choice: source:'T seq -> 'T
-val choiceWith: random:Random -> source:'T seq -> 'T
+val randomChoice: source:'T seq -> 'T
+val randomChoiceWith: random:Random -> source:'T seq -> 'T
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -81,7 +81,7 @@ val choiceWith: random:Random -> source:'T seq -> 'T
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.choice // "Charlie"
+let round1Order = allPlayers |> List.randomChoice // "Charlie"
 ```
 
 ### Choices
@@ -92,14 +92,14 @@ Two functions should be added to each module.
 
 ```fsharp
 // Array module
-val choices: count:int -> array:'T[] -> 'T[]
-val choicesWith: random:Random -> count:int -> array:'T[] -> 'T[]
+val randomChoices: count:int -> array:'T[] -> 'T[]
+val randomChoicesWith: random:Random -> count:int -> array:'T[] -> 'T[]
 // List module
-val choices: count:int -> list:'T list -> 'T list
-val choicesWith: random:Random -> count:int -> list:'T list -> 'T list
+val randomChoices: count:int -> list:'T list -> 'T list
+val randomChoicesWith: random:Random -> count:int -> list:'T list -> 'T list
 // Seq module
-val choices: count:int -> source:'T seq -> 'T seq
-val choicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
+val randomChoices: count:int -> source:'T seq -> 'T seq
+val randomChoicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -110,7 +110,7 @@ val choicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.choices 3 // ["Bob", "Dave", "Bob"]
+let round1Order = allPlayers |> List.randomChoices 3 // ["Bob", "Dave", "Bob"]
 ```
 
 ### Sample
@@ -121,14 +121,14 @@ Two functions should be added to each module.
 
 ```fsharp
 // Array module
-val sample: count:int -> array:'T[] -> 'T[]
-val sampleWith: random:Random -> count:int -> array:'T[] -> 'T[]
+val randomSample: count:int -> array:'T[] -> 'T[]
+val randomSampleWith: random:Random -> count:int -> array:'T[] -> 'T[]
 // List module
-val sample: count:int -> list:'T list -> 'T list
-val sampleWith: random:Random -> count:int -> list:'T list -> 'T list
+val randomSample: count:int -> list:'T list -> 'T list
+val randomSampleWith: random:Random -> count:int -> list:'T list -> 'T list
 // Seq module
-val sample: count:int -> source:'T seq -> 'T seq
-val sampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
+val randomSample: count:int -> source:'T seq -> 'T seq
+val randomSampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -139,7 +139,7 @@ val sampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.sample 3 // ["Charlie", "Dave", "Alice"]
+let round1Order = allPlayers |> List.randomSample 3 // ["Charlie", "Dave", "Alice"]
 ```
 
 # Drawbacks
@@ -199,8 +199,8 @@ N/A
 # Unresolved questions
 
 It's unclear, if more sophisticated overloads should be added:
- - Weights parameter for choices function
- - Counts parameter for sample function
+ - Weights parameter for randomChoices function
+ - Counts parameter for randomSample function
 
 ~~In .NET 6 and higher [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared) is available, but as soon as F# Core only targets standard, it can't use it. Is targeting higher .NET possible (not just for this feature, maybe some others need it)?
 Same question about new [.NET 8 apis](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), they could be reused in theory~~
--- a/RFCs/FS-1135-random-functions-for-collections.md
+++ b/RFCs/FS-1135-random-functions-for-collections.md
@@ -38,8 +38,15 @@ Shuffle function should returned a new shuffled collection of the same collectio
 Two functions should be added to each module.
 
 ```fsharp
-val shuffle: CollectionType<'T> -> CollectionType<'T>
-val shuffleRand: Random -> CollectionType<'T> -> CollectionType<'T>
+// Array module
+val shuffle: 'T[] -> 'T[]
+val shuffleRand: Random -> 'T[] -> 'T[]
+// List module
+val shuffle: 'T list -> 'T list
+val shuffleRand: Random -> 'T list -> 'T list
+// Seq module
+val shuffle: 'T seq -> 'T seq
+val shuffleRand: Random -> 'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -49,7 +56,6 @@ let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
 let round1Order = allPlayers |> List.shuffle // [ "Charlie"; "Dave"; "Alice"; "Bob" ]
 ```
 
-
 ### Choice
 
 Choice function should returned a random element from the collection.
@@ -57,8 +63,15 @@ Choice function should returned a random element from the collection.
 Two functions should be added to each module.
 
 ```fsharp
-val choice: CollectionType<'T> -> 'T
-val choiceRand: Random -> CollectionType<'T> -> 'T
+// Array module
+val choice: 'T[] -> 'T
+val choiceRand: Random -> 'T[] -> 'T
+// List module
+val choice: 'T list -> 'T
+val choiceRand: Random -> 'T list -> 'T
+// Seq module
+val choice: 'T seq -> 'T
+val choiceRand: Random -> 'T seq -> 'T
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -70,7 +83,6 @@ let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
 let round1Order = allPlayers |> List.choice // "Charlie"
 ```
 
-
 ### Choices
 
 Choices should select N elements from input collection in random order, once element is taken it can be selected again.
@@ -78,12 +90,21 @@ Choices should select N elements from input collection in random order, once ele
 Two functions should be added to each module.
 
 ```fsharp
-val choices: int -> CollectionType<'T> -> CollectionType<'T>
-val choicesRand: Random -> int -> CollectionType<'T> -> CollectionType<'T>
+// Array module
+val choices: int -> 'T[] -> 'T[]
+val choicesRand: Random -> int -> 'T[] -> 'T[]
+// List module
+val choices: int -> 'T list -> 'T list
+val choicesRand: Random -> int -> 'T list -> 'T list
+// Seq module
+val choices: int -> 'T seq -> 'T seq
+val choicesRand: Random -> int -> 'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
 [ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) should be raised if N is negative
+
+[InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception) should be raised if collection is empty
 
 Example:
 ```fsharp
@@ -98,12 +119,21 @@ Sample should select N elements from input collection in random order, once elem
 Two functions should be added to each module.
 
 ```fsharp
-val sample: int -> CollectionType<'T> -> CollectionType<'T>
-val sampleRand: Random -> int -> CollectionType<'T> -> CollectionType<'T>
+// Array module
+val sample: int -> 'T[] -> 'T[]
+val sampleRand: Random -> int -> 'T[] -> 'T[]
+// List module
+val sample: int -> 'T list -> 'T list
+val sampleRand: Random -> int -> 'T list -> 'T list
+// Seq module
+val sample: int -> 'T seq -> 'T seq
+val sampleRand: Random -> int -> 'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
 [ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) should be raised if N is greater than collection length or negative
+
+[InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception) should be raised if collection is empty
 
 Example:
 ```fsharp
@@ -171,5 +201,5 @@ It's unclear, if more sophisticated overloads should be added:
  - Weights parameter for choices function
  - Counts parameter for sample function
 
-In .NET 6 and higher [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared) is available, but as soon as F# Core only targets standard, it can't use it. Is targeting higher .NET possible (not just for this feature, maybe some others need it)?
-Same question about new [.NET 8 apis](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), they could be reused in theory
+~~In .NET 6 and higher [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared) is available, but as soon as F# Core only targets standard, it can't use it. Is targeting higher .NET possible (not just for this feature, maybe some others need it)?
+Same question about new [.NET 8 apis](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), they could be reused in theory~~
--- a/RFCs/FS-1135-random-functions-for-collections.md
+++ b/RFCs/FS-1135-random-functions-for-collections.md
@@ -195,4 +195,4 @@ N/A
 
 # Unresolved questions
 
-N/A
+[Additional function suggested](https://github.com/fsharp/fslang-design/discussions/731#discussioncomment-9328918)
--- a/RFCs/FS-1135-random-functions-for-collections.md
+++ b/RFCs/FS-1135-random-functions-for-collections.md
@@ -29,7 +29,6 @@ The following general rules are applied to all functions
  - New functions should be implemented in List, Array, Seq modules
  - All functions should have a variant with a [Random](https://learn.microsoft.com/en-us/dotnet/api/system.random) parameter
  - Shared thread-safe Random instance should be used for all basic functions
- - Functions will placed in Random submodule similar to Array.Parallel design
 
 ### Shuffle
 
@@ -38,24 +37,24 @@ Shuffle function should returned a new shuffled collection of the same collectio
 Two functions should be added to each module.
 
 ```fsharp
-// Array.Random module
-val shuffle: array:'T[] -> 'T[]
-val shuffleWith: random:Random -> array:'T[] -> 'T[]
-val shuffleInPlace: array:'T[] -> 'T[]
-val shuffleInPlaceWith: random:Random -> array:'T[] -> 'T[]
-// List.Random module
-val shuffle: list:'T list -> 'T list
-val shuffleWith: random:Random -> list:'T list -> 'T list
-// Seq.Random module
-val shuffle: source:'T seq -> 'T seq
-val shuffleWith: random:Random -> source:'T seq -> 'T seq
+// Array module
+val randomShuffle: array:'T[] -> 'T[]
+val randomShuffleWith: random:Random -> array:'T[] -> 'T[]
+val randomShuffleInPlace: array:'T[] -> 'T[]
+val randomShuffleInPlaceWith: random:Random -> array:'T[] -> 'T[]
+// List module
+val randomShuffle: list:'T list -> 'T list
+val randomShuffleWith: random:Random -> list:'T list -> 'T list
+// Seq module
+val randomShuffle: source:'T seq -> 'T seq
+val randomShuffleWith: random:Random -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.Random.shuffle // [ "Charlie"; "Dave"; "Alice"; "Bob" ]
+let round1Order = allPlayers |> List.randomShuffle // [ "Charlie"; "Dave"; "Alice"; "Bob" ]
 ```
 
 ### Choice
@@ -65,15 +64,15 @@ Choice function should returned a random element from the collection.
 Two functions should be added to each module.
 
 ```fsharp
-// Array.Random module
-val choice: array:'T[] -> 'T
-val choiceWith: random:Random -> array:'T[] -> 'T
-// List.Random module
-val choice: list:'T list -> 'T
-val choiceWith: random:Random -> list:'T list -> 'T
-// Seq.Random module
-val choice: source:'T seq -> 'T
-val choiceWith: random:Random -> source:'T seq -> 'T
+// Array module
+val randomChoice: array:'T[] -> 'T
+val randomChoiceWith: random:Random -> array:'T[] -> 'T
+// List module
+val randomChoice: list:'T list -> 'T
+val randomChoiceWith: random:Random -> list:'T list -> 'T
+// Seq module
+val randomChoice: source:'T seq -> 'T
+val randomChoiceWith: random:Random -> source:'T seq -> 'T
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -82,7 +81,7 @@ val choiceWith: random:Random -> source:'T seq -> 'T
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.Random.choice // "Charlie"
+let round1Order = allPlayers |> List.randomChoice // "Charlie"
 ```
 
 ### Choices
@@ -92,15 +91,15 @@ Choices should select N elements from input collection in random order, once ele
 Two functions should be added to each module.
 
 ```fsharp
-// Array.Random module
-val choices: count:int -> array:'T[] -> 'T[]
-val choicesWith: random:Random -> count:int -> array:'T[] -> 'T[]
-// List.Random module
-val choices: count:int -> list:'T list -> 'T list
-val choicesWith: random:Random -> count:int -> list:'T list -> 'T list
-// Seq.Random module
-val choices: count:int -> source:'T seq -> 'T seq
-val choicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
+// Array module
+val randomChoices: count:int -> array:'T[] -> 'T[]
+val randomChoicesWith: random:Random -> count:int -> array:'T[] -> 'T[]
+// List module
+val randomChoices: count:int -> list:'T list -> 'T list
+val randomChoicesWith: random:Random -> count:int -> list:'T list -> 'T list
+// Seq module
+val randomChoices: count:int -> source:'T seq -> 'T seq
+val randomChoicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -111,7 +110,7 @@ val choicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.Random.choices 3 // ["Bob", "Dave", "Bob"]
+let round1Order = allPlayers |> List.randomChoices 3 // ["Bob", "Dave", "Bob"]
 ```
 
 ### Sample
@@ -121,15 +120,15 @@ Sample should select N elements from input collection in random order, once elem
 Two functions should be added to each module.
 
 ```fsharp
-// Array.Random module
-val sample: count:int -> array:'T[] -> 'T[]
-val sampleWith: random:Random -> count:int -> array:'T[] -> 'T[]
-// List.Random module
-val sample: count:int -> list:'T list -> 'T list
-val sampleWith: random:Random -> count:int -> list:'T list -> 'T list
-// Seq.Random module
-val sample: count:int -> source:'T seq -> 'T seq
-val sampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
+// Array module
+val randomSample: count:int -> array:'T[] -> 'T[]
+val randomSampleWith: random:Random -> count:int -> array:'T[] -> 'T[]
+// List module
+val randomSample: count:int -> list:'T list -> 'T list
+val randomSampleWith: random:Random -> count:int -> list:'T list -> 'T list
+// Seq module
+val randomSample: count:int -> source:'T seq -> 'T seq
+val randomSampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
 
@@ -140,7 +139,7 @@ val sampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
 Example:
 ```fsharp
 let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
-let round1Order = allPlayers |> List.Random.sample 3 // ["Charlie", "Dave", "Alice"]
+let round1Order = allPlayers |> List.randomSample 3 // ["Charlie", "Dave", "Alice"]
 ```
 
 # Drawbacks
@@ -200,8 +199,8 @@ N/A
 # Unresolved questions
 
 It's unclear, if more sophisticated overloads should be added:
- - Weights parameter for choices function
- - Counts parameter for sample function
+ - Weights parameter for randomChoices function
+ - Counts parameter for randomSample function
 
 ~~In .NET 6 and higher [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared) is available, but as soon as F# Core only targets standard, it can't use it. Is targeting higher .NET possible (not just for this feature, maybe some others need it)?
 Same question about new [.NET 8 apis](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), they could be reused in theory~~
--- a/RFCs/FS-1135-random-functions-for-collections.md
+++ b/RFCs/FS-1135-random-functions-for-collections.md
@@ -153,7 +153,7 @@ Use online snippets, or provide a nuget package.
 
 Please address all necessary compatibility questions:
 
-* Is this a breaking change? **No, unless user defined those extensions himself**
+* Is this a breaking change? **No, unless user defined those extensions themselves**
 * What happens when previous versions of the F# compiler encounter this design addition as source code? **Should be fine**
 * What happens when previous versions of the F# compiler encounter this design addition in compiled binaries? **Should be fine**
 * If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct? **Will work as usual**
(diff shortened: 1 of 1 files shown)


RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Shuffle
+
+Shuffle function should returned a new shuffled collection of the same collection type.
+
+Two functions should be added to each module.
+
+```fsharp
+val shuffle: CollectionType<'T> -> CollectionType<'T>
+val shuffleRand: Random -> CollectionType<'T> -> CollectionType<'T>
```
cartermp, line 42: What is CollectionType?

As per the issue these are supposed to sit on the major F# collection types. I'd like to see a proposed signature for each.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,175 @@
...
+Choices should select N elements from input collection in random order, once element is taken it can be selected again.
+
+Two functions should be added to each module.
+
+```fsharp
+val choices: int -> CollectionType<'T> -> CollectionType<'T>
+val choicesRand: Random -> int -> CollectionType<'T> -> CollectionType<'T>
+```
+[ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
+
```
cartermp, line 85: What about if the collection is empty?

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,175 @@
...
+Sample should select N elements from input collection in random order, once element is taken it won't be selected again. N can't be greater than collection length
+
+Two functions should be added to each module.
+
+```fsharp
+val sample: int -> CollectionType<'T> -> CollectionType<'T>
+val sampleRand: Random -> int -> CollectionType<'T> -> CollectionType<'T>
+```
+[ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
+
```
cartermp, line 105: What if the collection is empty?

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,175 @@
...
+Example:
+```fsharp
+let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
+let round1Order = allPlayers |> List.sample 3 // ["Charlie", "Dave", "Alice"]
+```
+
+# Drawbacks
+
+System.Random interface has added some new methods in [.NET 8](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), where naming is a bit different. More new methods can eventually be added in future .NET versions.
```
cartermp, line 116: Ehhh, I think it's pretty unlikely that we could add these in. Maybe in the far, far future we can expect .NET 8 as a base dependency for FSharp.Core, but that's so far away that it's not even worth considering.

cartermp (changes requested):
Looks good overall, some changes needed though.

RFCs/FS-1135-random-functions-for-collections.md
vzarytovskii: Yeah, +1 we should use/target only what's available in ns2.0 at this point.

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: I've marked this unresolved question as resolved in the end of the RFC. The drawback itself I think should stay, since users will be choosing - to use .NET shuffling and sampling or F# shuffling and sampling

cartermp (approved):

T-Gro:
1. The APIs which can fail on non-null values ( choice,choices,sample with collections that are emty/too small) -> I would favour having a try* variant of them, to avoid developers doing their own wrappers around them to prevent exceptions.
  - A try- alternative is  in line with existing collections functions
  - 
2. (personally, this can also wait for an impl PR, but for me it is an important detail):
How do you plan to work with "Shared thread-safe Random instance ". Locking of a single global System.Random instance for the duration of entire operation? Or having multiple instance, either per-thread or in a pool?

3. Since games and simulations were mentioned, it would be good to thing of high perf scenarios:
- Just like it exists with sorting, could the Array.shuffle have an accompanying Array.shuffleInPlace to reduce copies?
- Especially with a more secure random generator, it could be handy to pre-generate the random indexes for shuffle in parallel (to live in Array.Parallel module) 
- 
4. The "with Random parameter" variants of the API : Does it only allow System.Random, or is https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator?view=net-7.0 considered?
(an abstraction covering both regular System.Random as well as RandonNumberGenerator could be requiring a random generating function as an argument).

Lanayx:
@T-Gro Hi, thank you for the review
1) Not all of collection functions have try variant, only the once where it makes sense. For example: max, average, reduce - just throw and that is ok for me. If function implies that collection is not empty, it is ok to throw, and all random functions imply that.
2) The second option, you can see my implementation here https://github.com/Lanayx/fsharp/blob/master/src/FSharp.Core/Random.fs
3) Agree, inplace shuffling can be useful, I'll add that
4) System.Random is not sealed and it doesn't implement RandomNumberGenerator, so if someone wants use the latter he can inherit System.Random and implement virtual methods (BTW System.Random.Shared does exactly this)

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+### Shuffle
+
+Shuffle function should returned a new shuffled collection of the same collection type.
+
+Two functions should be added to each module.
+
+```fsharp
+// Array module
+val shuffle: 'T[] -> 'T[]
+val shuffleRand: Random -> 'T[] -> 'T[]
```
dsyme, line 42: Suggest `shuffleWith` for this.  Although `With` often means `take a function` the `Randome` object effectively acts as the function.

Or else we could actually take a function parameter at this point, to make it more general, what do you think?

dsyme:
* I'm concerned about naming being confusing when colliding with existing functions, esp. `List.choices` v. `List.choose`. Perhaps

  * List.randomShuffle
  * List.randomShuffleWith
  * List.randomChoices
  * List.randomChoicesWith
  * List.randomSample
  * List.randomSampleWith

  It feels like the use of these functions is rare enough that adding "random" does help clarity here without too much cost.

* Please list naming of equivalent python functions

* Can we get any stats on use of equivalent python functions

* I'm concerned there's no weighted choice option (as mentioned in the unresolved questions)

* I'm concerned there are no functions to choose a single element.

* I'm bit concerned by having both "randomChoices" and "randomSample" without the naming aligning, but let's first take a look at the Python naming, also perhaps other languages.

Lanayx:
@dsyme thank you for the comments
1) I don't like the random prefix, it makes me feel like those functions shouldn't belong there at all, while for example `List.shuffle` feels nice like it has always been there. If you are strongly against extending List module - we can consider creating a separate module Random, so the code will look like
```fsharp
[1; 2; 3] |> Random.shuffle
[|1; 2; 3|] |> Random.shuffle
```
This will allow us to use overloads, but prevent using partial application. On one hand it is ok, on the other - it is not what the other collection functions look like, so it's not that consistent.

2) Passing Random seems to fit better than function, since I think we should make it easy to pass new Random.Shared around. Still using `shuffleWith` instead of `shuffleRand` feels good with me.
3) Idk about python functions usage, but I'm sure that every machine learning program should include random to shuffle data during several epochs.

4) Python: `shuffle`, `choice`, `choices`, `sample` (currently all names are taken from there). Java, Scala: `shuffle` only. Rust: `choose`, `choose_multiple`, `choose_weighted`, `shuffle`, `partial_shuffle`, `choose_multiple_fill`, `choose_multiple_weighted`. underscorejs: `shuffle`, `sample`

5) `choice` is the function to get single element

6) As for weights and counts parameters - my feeling is that while they are more specific to data science, so if somebody needs them - they can contribute later, so the the functions will be for example `List.choiceWeights` and 'List.sampleCounts'

7) I understand concern about choose and choice wording, but that's a minor issue for me. There are much bigger intersections in F# with .Net base types, like `Array` or `String` being a module and a class, so functions are intersected with .net methods. Or having ...Async methods in different libraries returning `Task` and not `Async` etc

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Changed to With

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Choices
+
+Choices should select N elements from input collection in random order, once element is taken it can be selected again.
+
+Two functions should be added to each module.
+
+```fsharp
+// Array module
+val choices: int -> 'T[] -> 'T[]
```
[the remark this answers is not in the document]
Lanayx: Done

vzarytovskii:
This looks fine to me overall.

I don't think we need a separate `Random` module for these functions, it will be a bit confusing on how it's related to collections.

As for naming - I don't have a strong opinion on that, since I don't have much experience in using those functions, but I share @dsyme's concern about naming - once they're in, there's no way for us changing or removing them in case they're confusing or lack some functionality. 

I personally think, that it's better to align naming and signatures to what existing functions in .NET and Python have.

Lanayx:
@dsyme I created a voting in [the discussion](https://github.com/fsharp/fslang-design/discussions/731) and it seems that people like the idea of prefix, so I changed all the names in PR

Lanayx:
Due to [the comment](https://github.com/fsharp/fslang-design/discussions/731#discussioncomment-5272327) I've changed prefixes to submodule, since this is already an implemented design in `Array.Parallel`

dsyme:
@Lanayx Sorry for the churn, but I don't think the module is the right way to go - could you revert to the prefix please?

Thanks!

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+Example:
+```fsharp
+let allPlayers = [ "Alice"; "Bob"; "Charlie"; "Dave" ]
+let round1Order = allPlayers |> List.randomShuffle // [ "Charlie"; "Dave"; "Alice"; "Bob" ]
+```
+
+### Choice
+
+Choice function should returned a random element from the collection.
```
abelbraaksma, line 62: nit:
```suggestion
The choice functions return a single random element from the given collection. The random choice is weighted evenly on the size of the collection.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+Two functions should be added to each module.
```
abelbraaksma, line 64: nit: (be careful with using "should" in RFC texts)
```suggestion
Two functions will be added to each module.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+val randomChoice: array:'T[] -> 'T
+val randomChoiceWith: random:Random -> array:'T[] -> 'T
+// List module
+val randomChoice: list:'T list -> 'T
+val randomChoiceWith: random:Random -> list:'T list -> 'T
+// Seq module
+val randomChoice: source:'T seq -> 'T
+val randomChoiceWith: random:Random -> source:'T seq -> 'T
+```
+[ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
```
abelbraaksma, line 77: And if the `Random` instance is `null` (depending on decision taken in #731)
```suggestion
[`ArgumentNullException`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) is raised if collection argument is `null`, or if the `random` argument is `null`.
```

RFCs/FS-1135-random-functions-for-collections.md
abelbraaksma: We might want to add a note about the direction of the implementation here. People may assume that it isn't necessary to traverse the whole sequence or list. However, since the choice should be balanced in relation to the size of the list or sequence, we will need to do a full traversal.

In the case of lists and sequences, that means that we (probably) need to create a temporary local copy of the full sequence. The alternative is multiple traversals.

RFCs/FS-1135-random-functions-for-collections.md
abelbraaksma: I think what you mean to say here is that we're adding on top of an existing surface area of methods in .NET. I don't think you propose here to use, or refer to these new methods? (they are not part of NetStandard).

However, that said, I'm not opposed mentioning it, but maybe more along the lines of:

```suggestion
Users may be tempted to use some of the recently added method of `System.Random` that also apply to collections instead of the ones we add in FSharp.Core. It may also be confusing to some, especially since the naming over there is slightly different. See [.NET 8 What's New](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness).
```

It's certainly not an unresolved question ;).

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+let round1Order = allPlayers |> List.randomSample 3 // ["Charlie", "Dave", "Alice"]
+```
+
+# Drawbacks
+
+System.Random interface has added some new methods in [.NET 8](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), where naming is a bit different. More new methods can eventually be added in future .NET versions.
+
+# Alternatives
+
+Use online snippets, or provide a nuget package.
```
abelbraaksma, line 151: It's creative of you to think of this as an alternative :). Of course, anybody could just provide their own methods (trivial enough to do).

Usually, for RFC's like these that add functions to the surface area, we just say:

```suggestion
Not doing this.
```

as there really isn't a good alternative :).

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+# Compatibility
+
+Please address all necessary compatibility questions:
```
abelbraaksma, line 155: nit (is from the template, can be removed)
```suggestion
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+* Is this a breaking change? **No, unless user defined those extensions themselves**
```
abelbraaksma, line 157: nit:
```suggestion
* Is this a breaking change? **No**
```

User-defined implementations have precedence. So even if, this is not going to be a problem. Users can *always* implement their own version of any of the FSharp.Core functions.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+* What happens when previous versions of the F# compiler encounter this design addition as source code? **Should be fine**
```
abelbraaksma, line 158:
```suggestion
* What happens when previous versions of the F# compiler encounter this design addition as source code? **Library function, not applicable**
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries? **Should be fine**
```
abelbraaksma, line 159:
```suggestion
* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries? **Library function, not applicable***
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+* If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct? **Will work as usual**
+
+# Pragmatics
+
+## Diagnostics
+
+Please list the reasonable expectations for diagnostics for misuse of this feature. **I don't see a way to misuse it**
```
abelbraaksma, line 166: Not sure that goes in here, but any random implementation can be valid even if it always returns the exact same value. In that case, the same item would always be returned. But since this is essentially true of anything that anybody can write in their own code, it's probably not worth mentioning.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+    * Breakpoints/stepping
+    * Expression evaluator
+    * Data displays for locals and hover tips
+* Auto-complete
+* Tooltips
+* Navigation and Go To Definition
+* Colorization
+* Brace/parenthesis matching
+
+**Should work just like for other collections functions**
```
abelbraaksma, line 182: easy on the use of "should" ;)
```suggestion
**These will work just like for other collections functions**
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Performance
+
+Please list any notable concerns for impact on the performance of compilation and/or generated code
```
abelbraaksma, line 186: nit (part of template, can be removed)
```suggestion
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+* For existing code **Existing code should not be affected**
```
abelbraaksma, line 188: nit
```suggestion
* For existing code **Existing code won't be affected**
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+* For the new features **Performance should be respected when implementing this feature, since it can be used in performance-sensitive scenaris**
```
abelbraaksma, line 189: nit
```suggestion
* For the new features **Performance should be respected when implementing this feature, since it can be used in performance-sensitive scenarios**
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Scaling
+
+Algorithmic complexity of the new features should be O(n) or less.
```
abelbraaksma, line 193: We may need to specify this per function. Basically: O(n) for list and seq functions. O(1) for most array functions (except shuffle).

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+* For existing code **Existing code should not be affected**
+* For the new features **Performance should be respected when implementing this feature, since it can be used in performance-sensitive scenaris**
+
+## Scaling
+
+Algorithmic complexity of the new features should be O(n) or less. 
+
+## Culture-aware formatting/parsing
+
+N/A
```
abelbraaksma, line 197:
```suggestion
Each error, unless already defined in FSharp.Core, will need to be translated.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+# Unresolved questions
+
+It's unclear, if more sophisticated overloads should be added:
+ - Weights parameter for randomChoices function
+ - Counts parameter for randomSample function
```
abelbraaksma, line 203: You have the `count` argument already, this ca be dropped:
```suggestion
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+Algorithmic complexity of the new features should be O(n) or less. 
+
+## Culture-aware formatting/parsing
+
+N/A
+
+# Unresolved questions
+
+It's unclear, if more sophisticated overloads should be added:
+ - Weights parameter for randomChoices function
```
abelbraaksma, line 202: I think we can safely say that we won't do this in the current implementation? 

The only remaining unresolved question is the choice of randomizer argument: either `Random` or `unit - float`, see #731.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+ - Counts parameter for randomSample function
+
+~~In .NET 6 and higher [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared) is available, but as soon as F# Core only targets standard, it can't use it. Is targeting higher .NET possible (not just for this feature, maybe some others need it)?
```
abelbraaksma, line 205: No, not possible (not yet, at least). It would mean maintaining multiple versions of FSharp.Core and we are trying really, really hard to prevent that.

I think you an drop this as _unresolved question_. We will just cherry-pick from the existing implementation, as suggested above.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+Same question about new [.NET 8 apis](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), they could be reused in theory~~
```
abelbraaksma, line 206: Yes. But there will always be new versions and new interfaces, methods etc. If someone wants to use them, they can, by simply opening the relevant namespace.

I think you an drop this as _unresolved question_ too.

abelbraaksma (changes requested):
Most of the below suggestions are just some small nits, please have a look. I think we're quite near completion here!

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+ - Using F# for data science and machine learning (like building a neural network), where data shuffling plays important role
+ - Building games, where random sampling is used for generating random levels, random decks, etc.
+ - Building simulations, where random sampling is used for generating random input data
+
+# Detailed design
+
+### General
+
+The following general rules are applied to all functions
+ - New functions should be implemented in List, Array, Seq modules
```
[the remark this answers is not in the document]
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+ - All functions should have a variant with a [Random](https://learn.microsoft.com/en-us/dotnet/api/system.random) parameter
+ - Shared thread-safe Random instance should be used for all basic functions
```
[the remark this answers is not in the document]
Lanayx: Updated a bit differently not to mention exact implementation details

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Shuffle
+
+Shuffle function should returned a new shuffled collection of the same collection type.
```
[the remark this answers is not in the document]
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+
+Two functions should be added to each module.
```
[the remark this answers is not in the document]
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,206 @@
...
+val randomShuffleInPlace: array:'T[] -> 'T[]
+val randomShuffleInPlaceWith: random:Random -> array:'T[] -> 'T[]
+// List module
+val randomShuffle: list:'T list -> 'T list
+val randomShuffleWith: random:Random -> list:'T list -> 'T list
+// Seq module
+val randomShuffle: source:'T seq -> 'T seq
+val randomShuffleWith: random:Random -> source:'T seq -> 'T seq
+```
+[ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) should be raised if collection is null
```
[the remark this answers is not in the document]
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed text. As for implementation guidance, I don't think it's necessary, since doing random selection on only part of collection will look unnatural and easy to spot on review.

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: I don't know what to write here either :) So leaved it as is

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: As far as I remember from my implementation, there will be no new errors introduced

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: I've cleaned up this section at all for now. I will mention your suggestion later if we don't agree about it in the linked discussion thread within few days.

Lanayx: Ok, added link to your suggestion

RFCs/FS-1135-random-functions-for-collections.md
abelbraaksma: Good points, tx.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -179,28 +178,21 @@ Please list the reasonable expectations for tooling for this feature, including
...
 # Unresolved questions
 
-It's unclear, if more sophisticated overloads should be added:
- - Weights parameter for randomChoices function
- - Counts parameter for randomSample function
-
-~~In .NET 6 and higher [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared) is available, but as soon as F# Core only targets standard, it can't use it. Is targeting higher .NET possible (not just for this feature, maybe some others need it)?
-Same question about new [.NET 8 apis](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#methods-for-working-with-randomness), they could be reused in theory~~
\ No newline at end of file
+[Additional function suggested](https://github.com/fsharp/fslang-design/discussions/731#discussioncomment-9328918)
```
abelbraaksma, line 198: Thanks for linking the open point here. Let's resolve that soon(ish), so we can replace this simply with nothing, and I can merge this proposal for you, so we can start implementing it before the deadline of the next F# gets too close.

```suggestion
None
```

abelbraaksma (changes requested):

abelbraaksma:
After we've resolved the unresolved question, I'll give this RFC one more quick review and then I think I can merge it for you.

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -29,6 +29,7 @@ This feature is motivated by the following use cases:
 The following general rules are applied to all functions
  - New functions should be implemented in `List`, `Array`, `Seq` modules
  - Each function should have a variant that takes a [Random](https://learn.microsoft.com/en-us/dotnet/api/system.random) argument
+ - Each function should have a variant that takes a custom `randomizer` function. This function should return a `float` value from `0..1` range sampled from the desired distribution.
```
abelbraaksma, line 32: nit, let's be precise:
```suggestion
 - Each function should have a variant that takes a custom `randomizer` function. This function should return a `float` that is greater than or equal to `0.0`, and less than `1.0`, like [Random.NextDouble](https://learn.microsoft.com/en-us/dotnet/api/system.random.nextdouble?view=net-8.0)).
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -68,17 +75,22 @@ The following functions will be added to each module.
...
 // Seq module
 val randomChoice: source:'T seq -> 'T
 val randomChoiceWith: random:Random -> source:'T seq -> 'T
+val randomChoiceBy: randomizer: (unit -> float) -> source:'T seq -> 'T
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) is raised if collection is `null`, or if the `random` argument is `null`.
 
 [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception) is raised if collection is empty.
 
+[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if randomizer returns a value outside the `0..1` range.
```
abelbraaksma, line 92: nit, let's be precise:
```suggestion
[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if the `randomizer` function returns a `float` that is less than `0.0` or greater or equal to `1.0`.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -95,19 +107,24 @@ The following functions will be added to each module.
...
 val randomChoicesWith: random:Random -> count:int -> source:'T seq -> 'T seq
+val randomChoicesBy: randomizer: (unit -> float) -> count:int -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) is raised if collection is `null`, or if the `random` argument is `null`.
 
 [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if N is negative.
 
 [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception) is raised if collection is empty.
 
+[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if randomizer returns a value outside the `0..1` range.
```
abelbraaksma, line 126: nit, let's be precise:
```suggestion
[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if the `randomizer` function returns a `float` that is less than `0.0` or greater or equal to `1.0`.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -124,19 +141,23 @@ The following functions will be added to each module.
...
 val randomSampleWith: random:Random -> count:int -> source:'T seq -> 'T seq
+val randomSampleBy: randomizer: (unit -> float) -> count:int -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) is raised if collection is `null`, or if the `random` argument is `null`.
 
 [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if N is greater than collection length or is negative.
 
 [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception) is raised if collection is empty.
 
+[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if randomizer returns a value outside the `0..1` range.
```
abelbraaksma, line 159: nit, let's be precise:
```suggestion
[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if the `randomizer` function returns a `float` that is less than `0.0` or greater or equal to `1.0`.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -41,17 +42,23 @@ The following functions will be added to each module.
...
 val randomShuffleWith: random:Random -> list:'T list -> 'T list
+val randomShuffleBy: randomizer: (unit -> float) -> list:'T list -> 'T list
 // Seq module
 val randomShuffle: source:'T seq -> 'T seq
 val randomShuffleWith: random:Random -> source:'T seq -> 'T seq
+val randomShuffleBy: randomizer: (unit -> float) -> source:'T seq -> 'T seq
 ```
 [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) is raised if collection is `null`, or if the `random` argument is `null`.
 
+[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if randomizer returns a value outside the `0..1` range.
```
abelbraaksma, line 60: nit, let's be precise:
```suggestion
[ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception) is raised if the `randomizer` function returns a `float` that is less than `0.0` or greater or equal to `1.0`.
```

abelbraaksma (changes requested):
This is great, thanks! A few minor nits, mainly about being precise (and follow the language of `NextDouble`).

I think this is essentially ready. While I did mention a few things in the discussion thread, up to you to mention them here as well (i.e. on infinite sequences (note that .NET only supports arrays and spans, in contrast), and on the necessary caching).

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
abelbraaksma: Hey @Lanayx, it appears we both forgot about this line. We shouldn't leave the remnants of the template hanging around (I know, it happens, but we shouldn't ;) ).

This is my suggestion:

```suggestion
There are no known diagnostics on any abuse or misuse of this feature.
```

RFCs/FS-1135-random-functions-for-collections.md
```diff
@@ -0,0 +1,219 @@
...
+    * Breakpoints/stepping
+    * Expression evaluator
+    * Data displays for locals and hover tips
+* Auto-complete
+* Tooltips
+* Navigation and Go To Definition
+* Colorization
+* Brace/parenthesis matching
+
+**These will work just like for other collections functions**
```
abelbraaksma, line 202:
```suggestion
## Tooling

n/a, this is a F# Core library addition.

```

abelbraaksma (approved):
Approved, but see these two suggestions of text remnants from the RFC template (not a blocker):

RFCs/FS-1135-random-functions-for-collections.md
Lanayx: Fixed

RFCs/FS-1135-random-functions-for-collections.md
abelbraaksma: This has since been resolved

abelbraaksma:
@dsyme and @vzarytovskii, this proposal is ready. If you can give it your thumbs up, I can merge it (or you can ;) ).

dsyme (approved):

dsyme:
Merged! Thank you everyone for your hard work on this

KevinRansom:
@dsyme -- please consider the revoking the approval for this suggestion.

FSharp.Core is already too big, and so far all attempts to refactor it to make it smaller have failed.  Useful helper functions such as these belong in an external library where they can be selected by a developer who is working on scenarios where this type of functionality is useful.

We currently ship fsharp.core as a nuget package embedded within the dotnet sdk:

- The current FSharp.Core.dll is currently 3MB and it looks like when nullability is supported it will bloat by perhaps an additional .5 MB, because we will have to duplicate the metadata stream for nullable aware compilers and legacy compilers.  From time to time we get pressure from the SDK team to become an optional install feature because of our size and relatively small user base.  I think whenever they hear a customer complaining that VMs for the dotnet SDK have to be larger than for other cloud platforms and languages.

In general we have only added APIs to FSharp.Core when they were widely applicable and supported an idiomatic programming style, or interop with C# assemblies of course some that don't match these criteria have slipped in.

But these suggested APIs deserve to live in a support library rather than FSharp.Core, someone should work on it.  Because we have many gaps with the Numpy data library and FSharp.Core is not the place to address them.

smoothdeveloper:
FSharp.Stats?

Lanayx:
>  Because we have many gaps with the Numpy data library and FSharp.Core is not the place to address them.

Random functions are not part of Numpy, they are part of Python standard library, since they are very basic and widely applicable and were really "missing" in F# since inception
https://www.w3schools.com/python/module_random.asp

krauthaufen:
I too think that bloated libraries are hard to maintain and refactor since i have experience with it sadly. Maybe this would be a good starting point for a second nuget package?
Could live in this repo and just produce something called FSharp.Collections maybe?
I personally think that randomozing collections is somewhat oddly specific for a standard library.
Just my thoughts

Lanayx:
> bloated libraries are hard to maintain and refactor

Fully agree. However we should differentiate that from absence of the very basic and generally applicable functionality.

> Maybe this would be a good starting point for a second nuget package?

I don't think so

> I personally think that randomozing collections is somewhat oddly specific for a standard library.

The presence of random functions in Python standard library is one of the many things that lowered the barrier of ML entry for newcomers and it very positively influenced language popularity, we should follow the success path here rather than avoiding it.

dsyme:
Ultimately the team looking after the `dotnet/fsharp` repo do really have a right of veto on this. 

@KevinRansom I'd request that we get those concerns raised much, much earlier in the design process. e.g. at the suggestion stage, or minimally at the RFC-discussion stage.  This RFC PR was open 12 months.

@Lanayx Given @KevinRansom's concerns could we get a measure of the size delta?

@Lanayx I would favour dropping the `With` variants if it helps with size concerns. I'm ambivalent about those anyway. @Lanayx is there any technical reason to include them, e.g. is passing `rnd.Next` to the `By` variant equivalent?

Lanayx:
@dsyme As for size question, here is measurements
Current main: **3261** KB
With random changes: *3293* KB
So the change is 32KB, which is 1% size change. As for dropping `.With` method I tend to disagree, since it's the fastest option, 2.5 times faster than `.By` option

```fsharp
module TestData =
    let arr = Array.init 1000 id

[<MemoryDiagnoser>]
type Rand() =


    [<Benchmark(Baseline = true)>]
    member _.Bcl () =
        let newArr = Array.copy TestData.arr
        Random.Shared.Shuffle newArr
        newArr

    [<Benchmark>]
    member _.RandomWith () =
        TestData.arr |> Array.randomShuffleWith Random.Shared

    [<Benchmark>]
    member _.RandomBy () =
        TestData.arr |> Array.randomShuffleBy Random.Shared.NextDouble
```

```
| Method     | Mean      | Error     | StdDev    | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|----------- |----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
| Bcl        |  4.229 us | 0.0650 us | 0.0608 us |  1.00 |    0.00 | 0.4730 |   3.93 KB |        1.00 |
| RandomWith |  3.743 us | 0.0354 us | 0.0331 us |  0.89 |    0.01 | 0.4768 |   3.93 KB |        1.00 |
| RandomBy   | 10.185 us | 0.0871 us | 0.0728 us |  2.41 |    0.04 | 0.4730 |   3.95 KB |        1.01 |
```
If 1% increase is indeed too much and I had to drop anything from API, I'd drop `.By` version of functions or `Seq` module functions (since all those functions internally do caching to the array as a first operation)

Lanayx:
@brianrourkeboll It will be slower anyway, because of checking value range, and because of doing extra calculations of converting float 0..1 value to int min..max value

brianrourkeboll:
@Lanayx Hmm, it seems to be possible to get `randomShuffleBy` much closer by making a few tweaks:

[Benchmark source](https://gist.github.com/brianrourkeboll/9f18cc71c821eef9cf9053cad34e702c)

```console
| Method | Mean     | Error     | StdDev    | Ratio | RatioSD |
|------- |---------:|----------:|----------:|------:|--------:|
| Bcl    | 3.386 us | 0.0312 us | 0.0276 us |  1.00 |    0.00 |
| Pr     | 6.721 us | 0.0795 us | 0.0705 us |  1.99 |    0.03 |
| Faster | 3.881 us | 0.0290 us | 0.0271 us |  1.15 |    0.01 |
```

The lambda does seem to be required if you want full inlining, though, i.e.,

```fsharp
Faster.randomShuffleBy Random.Shared.NextDouble TestData.arr
```

is slower than

```fsharp
Faster.randomShuffleBy (fun () -> Random.Shared.NextDouble ()) TestData.arr
```

dsyme:
Interesting thanks! BTW what's the cost of the locking involved in Random.Shared? Just out of curiosity really.

I think from this I still approve the RFC, and my preference would still be to remove the With variants (@Lanayx is there any reason besides perf to include them?) 

For me the deciding factor from a design perspective is that a large number of F# teaching scenarios become much simpler to teach if you just have `List.randomShuffle` and so on directly available, without having to explain System.Random or any .NET functionality (or the awful `Span<T>` overloads on `System.Random.Shared.Shuffle` - a new presence in the .NET world that is really problematic for teaching - imagine trying to explain to a beginner student why their Shuffle overload resolution fails!). 

That's really one of the main things FSharp.Core is for: to present a coherent, teachable, usable, portable programming model that captures most common in-memory programming scenarios before moving on to advanced data structures or interop with system features or UX or tensors and so on.  Doing basic random permutations has been a part of programming models like this ever since the days of Python and before. It is in Python for a reason, and should be in F# for the same reason.  The addition of System.Random.Shared is a factor in this - it indicated that .NET has embraced the same principle. Therefore F# should embrace the principle too from its own perspective. 

The owners of dotnet/fsharp still have a veto right on this one. Size is a factor affecting many scenarios that might not be obvious.

brianrourkeboll:
@dsyme

> BTW what's the cost of the locking involved in Random.Shared? Just out of curiosity really.

Do you mean [`System.Random.Shared` in the BCL](https://github.com/dotnet/runtime/blob/fb1fe7f3481fc88fa477268c18d60d824b83c314/src/libraries/System.Private.CoreLib/src/System/Random.cs#L54-L55)? If so, [it uses an implementation with a `ThreadStatic` private field for thread-safety](https://github.com/dotnet/runtime/blob/fb1fe7f3481fc88fa477268c18d60d824b83c314/src/libraries/System.Private.CoreLib/src/System/Random.cs#L359-L468) instead of explicit locking (that technique is also what lets it devirtualize and then inline calls to `Next`, etc.).

Compare: https://github.com/dotnet/fsharp/pull/17277#discussion_r1631486448 and https://github.com/dotnet/fsharp/pull/17277#discussion_r1631649968.

dsyme:
> Do you mean [System.Random.Shared in the BCL](https://github.com/dotnet/runtime/blob/fb1fe7f3481fc88fa477268c18d60d824b83c314/src/libraries/System.Private.CoreLib/src/System/Random.cs#L54-L55)? If so, [it uses an implementation with a ThreadStatic private field for thread-safety](https://github.com/dotnet/runtime/blob/fb1fe7f3481fc88fa477268c18d60d824b83c314/src/libraries/System.Private.CoreLib/src/System/Random.cs#L359-L468) instead of explicit locking (that technique is also what lets it devirtualize and then inline calls to Next, etc.).

Got it, thanks!

Lanayx:
> For me the deciding factor from a design perspective is that a large number of F# teaching scenarios become much simpler to teach if you just have List.randomShuffle and so on directly available, without having to explain System.Random or any .NET functionality

But `List.randomShuffle` and so on _will_ be directly available, so this probably doesn't apply? Very early beginners will just be able to use simplest version of random methods while studying. But anyway, `System.Random` is so central to random operations in .NET that even a beginner will have to learn it very soon. Also learners usually just copy documentation or samples or what ChatGPT tells them.

> I think from this I still approve the RFC, and my preference would still be to remove the With variants (@Lanayx is there any reason besides perf to include them?)

From my hands-on perspective ideal F# API that is based on C# experience should the following properties:
1) Simpler or easier to use than C# API
2) Same (or better) performance that C# API
3) Same (or better) functionality set

So I look at different options with regard to those properties:

- Simple versions (randomShuffle etc) 1. + 2. A bit worse that C# 3. Random characteristics will be worse than with Random.Shared
- .By version (randomShuffleBy etc) 1. A bit harder than C# (passing delegate is harder to comprehend than passing object) 2. Worse than C# (without hacks, with hacks still a bit worse) 3. +
- .With version (randomShuffleWith etc) 1. + 2. + 3. +

Based on above I would use the .With option in my work project (with Random.Shared) it was available. If it's not available, I'll have to add one more method to my own collection of missing language functions (near `isNotNull`) in all projects that need it. Also, if it's not available, what will we recommend to use by default in production? Will we tell to use this snippet:
```fsharp
Array.randomShuffleBy (fun () -> Random.Shared.NextDouble ()) TestData.arr
```
This would raise a thought in my mind - "They had a chance to make it nice after 20 years, but failed it". 

Ultimately the .With option would be not needed it Fsharp.Core targeted .NET 8, because the implementation could just be based on `Random.Shared` (which is by far the main case for .With option). But when (or if) this happens is unknown to me, so I'd keep it. Additional benefit is that there are many alternative implementations (like [here](https://numerics.mathdotnet.com/Random)) that all inherit `System.Random` and user can easily leverage one of them using .With versions without sacrificing performance.

krauthaufen:
Just a thought: How exactly is randomShuffleBy any different from sortBy?

Signature-wise it seems to be identical and also the semantics seem to be the same.

Lanayx:
> Just a thought: How exactly is randomShuffleBy any different from sortBy?

Shuffle algorithmic complexity is O(N)

krauthaufen:
So the generated double is actually interpreted as an index for classic swap-operations?

If so, i think it's very odd (what happens when i return NaN or simply 3.0, etc)

Lanayx:
> So the generated double is actually interpreted as an index for classic swap-operations like [this](https://github.com/aardvark-platform/aardvark.base/blob/05377715c397010129d8f6ed08ffcd218578d605/src/Aardvark.Base/Random/IRandomUniform.cs#L607)?

Right

> If so, i think it's very odd (what happens when i return NaN or simply 3.0, etc)

It's covered in the RFC, the `ArgumentOutOfRange` exception will be thrown

PR #734, opened 2023-04-20
Create FS-1135-map-binary-search.md

reinux:
[Rendered RFC](https://github.com/reinux/fslang-design/blob/patch-1/RFCs/FS-1135-map-binary-search.md)

### DIFF
--- a/RFCs/FS-1135-map-binary-search.md
+++ b/RFCs/FS-1135-map-binary-search.md
@@ -30,14 +30,12 @@ Currently, the only way to accomplish something similar is to iterate through `M
 
 # Detailed design
 
-This is the bulk of the RFC. Explain the design in enough detail for somebody familiar
-with the language to understand, and for somebody familiar with the compiler to implement.
-This should get into specifics and corner-cases, and include examples of how the feature is used.
-
 The function currently has the shape:
 ```fsharp
 Map.binarySearch : 'Key -> Map<'Key, 'Value> -> ('Key, 'Value) option * ('Key, 'Value) option * ('Key, 'Value) option
 ```
+
+The function is implemented in Map.fs in FSharp.Core. It should not throw any additional exceptions except `StackOverflowException` and `OutOfMemoryException`, including in the case where the Map is empty.
 
 Example use in a hypothetical Elmish/Feliz view:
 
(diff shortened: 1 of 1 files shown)


dsyme:
@KevinRansom @vzarytovskii @abelbraaksma Comments on this please?

Martin521:
I like the functionality and the API.
I am not sure about the name. "Binary search" is a well defined algorithm for arrays. Just Map.search? Map.tryFindWithNeighbors?

reinux:
> I like the functionality and the API. I am not sure about the name. "Binary search" is a well defined algorithm for arrays with a single return value. Just Map.search? Map.tryFindWithNeighbors?

The more I think about it, the more I'm liking `Map.search`. Even if you could validly argue that it's a special case of binary search (and I'm starting to have doubts), you probably wouldn't be the only one raising an eyebrow.

`search` is a nice, short name, and the return type, whatever we end up landing on, will do a fine job describing what it does anyway.


As for the return type, I think the only option we haven't explored yet is a single-case union with named parameters:

```fsharp
type MapSearchResult<'k, 'v> = MapSearchResult of left: ('k, 'v) option * exact: ('k, 'v) option * right: ('k, 'v) option
```

This would effectively be the same as the tuple version, except that we'd be trading conciseness with some IDE assistance.

I'm likewise unsure whether this is the best approach either.

T-Gro:
From the options so far, this API seems the best for avoiding errors and for communicating intent clearly.

```fsharp
type MapSearchResult<'k, 'v> =
  MapSearchResult of left: ('k, 'v) option * exact: ('k, 'v) option * right: ('k, 'v) option
```

`Map.search` is fitting

Happypig375:
The `MapSearchResult` return type seems like a workaround for the lack of [named tuple fields](https://github.com/fsharp/fslang-suggestions/issues/1434) in F#.

T-Gro:
There are more benefits to a custom type over tuples for API contracts:
- Binary stability in the case of more fields added in the future (whereas going from 3-tuple to e.g. 4-tuple is a binary breaking change)
- Understood by older compilers and/or IDE tools
- Can have extensions defined

Happypig375:
@T-Gro A binary breaking change would still occur when adding another field to `MapSearchResult` on the construction side, even though deconstruction, based on member access, wouldn't break.

The point about older compilers and IDE tools may apply but documentation would help - we can't be stuck in the past with API additions.

Also, are extensions really a point of concern? What are the use cases for extending MapSearchResult at all?

T-Gro (changes requested):
See https://github.com/fsharp/fslang-design/pull/734#issuecomment-2653758808 for the last changes to incorporate.
After that is done, this PR can be merged.

T-Gro:
@reinux :

Are you interested in pursuing the RFC and related impl further?
I will mark the PR here as a draft for now, but feel free to reopen it and ask for final review any time.

reinux:
I hope to get back to it soon! Swamped with multiple projects at the moment but I'll keep this tab open.

PR #735, opened 2023-04-26

dsyme:
We're looking at this again and after discussion with @vzarytovskii are looking more deeply at `string | null` as the syntax for nullable reference types.  

We've concluded `string?` overly normalizes nullable reference types as the canonical representation of optionality in F#

### DIFF
--- a/RFCs/FS-1060-nullable-reference-types.md
+++ b/RFCs/FS-1060-nullable-reference-types.md
@@ -18,9 +18,7 @@ Conceptually, reference types can be thought of as having two forms:
 * Normal reference types
 * Nullable reference types
 
-Warnings are emitted when reference types and nullable reference types are not used according with their intent, and these warnings are tunable as errors.
-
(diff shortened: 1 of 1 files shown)


charlesroddie:
It's a mistake to add a new syntax, different from other dotnet languages, for a feature designed specifically for interop with other dotnet languages. Programmers will now need to know both syntaxes to use this feature.

The RFC now has a syntax designed to be a bit awkward, while it also has sugar (like `match ... with null | x -> x.Method`) designed to save programmers typing time when using the feature extensively. That's a bit inconsistent.

vzarytovskii:
> It's a mistake to add a new syntax, different from other dotnet languages, for a feature designed specifically for interop with other dotnet languages. Programmers will now need to know both syntaxes to use this feature.
> 
> The RFC now has a syntax designed to be a bit awkward, while it also has sugar (like `match ... with null | x -> x.Method`) designed to save programmers typing time when using the feature extensively. That's a bit inconsistent.

F# also interopz with JS and TS, syntax also fits with literal types, which are under consideration by Don.

PR #742, opened 2023-07-18

smoothdeveloper:
removing a spooky statement around F# advancing in .NET (which is a given), and giving more wiggling room than "few years" for .NET landscape to "change".

The total set of C# codebases tend to evolve at another pace than the open source ecosystem of libraries or the BCL implementation.


NinoFloris:
In a few years has happened already though, the entire BCL is annotated, many libraries as well. So *updating* this document to mention medium term IMO doesn't quite fit.

smoothdeveloper:
@NinoFloris, fair, do you prefer the last edit, which takes the stance of looking back (evolution already took place)?

cartermp (approved):
I like the last edit. Mergin'

PR #750, opened 2023-09-07
Move RFCs to coresponding folders

vzarytovskii:
We have surprisingly much less RFCs than features in F# 8, am I missing something? I.e. I moved 5 RFCs to F# 8, but we have 24 language features enabled in compiler for version 8.


nojaf:
>  but we have 24 language features enabled in compiler for version 8.

Where can we find that list again?

vzarytovskii:
> > but we have 24 language features enabled in compiler for version 8.
> 
> Where can we find that list again?

https://github.com/dotnet/fsharp/blob/f48d509f0d6962d499cb795d346b9b196b779b4c/src/Compiler/Facilities/LanguageFeatures.fs#L160-L183

0101 (approved):

PR #753, opened 2023-09-23
initialize RFC-1137 Prefer extension method over property when arguments are provided

smoothdeveloper:

Just a place holder for now, in order to have all links setup (https://github.com/fsharp/fslang-design/discussions/752).

### DIFF
--- a/preview/FS-1137-prefer-extensions-method-over-property-when-arguments-provided.md
+++ b/preview/FS-1137-prefer-extensions-method-over-property-when-arguments-provided.md
@@ -1,0 +1,22 @@
+# F# RFC FS-1137 - Prefer extension method over intrinsic property when arguments are provided
+* [x] Approved in principle
+* [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/1039)
+* [x] Implementation: [In progress](https://github.com/dotnet/fsharp/pull/16032)
+* [x] [Discussion](https://github.com/fsharp/fslang-design/discussions/752)
+
+# Summary
+
+# Motivation
+
+# Detailed design
+
+## Implementation details
+
+## Documentation
+
+# Drawbacks
+
+# Alternatives
+
+# Unresolved questions
+


smoothdeveloper:
Is it possible to merge this one? I've added the exec summary and I'm referencing the RFC identifier in the PR already, just want to make sure 1137 is stabilized identifier for this.

psfinaki:
@dsyme @cartermp @baronfel - can you share your thoughts?

preview/FS-1137-prefer-extensions-method-over-property-when-arguments-provided.md
```diff
@@ -0,0 +1,115 @@
...
+* they can't bear the same name in the IL (properties have a `get_` or `set_` prefix addorned to the IL methods implementation of getter and setter)
+* they can be defined as extension methods in C# & VB.NET, and in F# with this feature, as method in type extension and extension methods
+
+It may make sense to lift this restriction in case C# moves in this direction.
+
+In meantime, attempting to define such intrinsic methods still results in 
+
+>  error FS0434: The property '{property name}' has the same name as a method in type '{type name}'.
+
+## FCS breaking change in items returned `DeclarationListInfo`
```
vzarytovskii, line 102: Do we know of any cases if this can be breaking change in runtime? e.g. if property exposes function or delegate?

```fsharp
open System.Runtime.CompilerServices
type TypeWithProperty () =
   member _.X
     with get () = fun (x:int) -> printfn $"Property: {x}"; x
[<Extension>]
type TypeWithPropertyExt =
     [<Extension>]
     static member X (t: TypeWithProperty, i: int) = printfn $"Method: {i}"; i
let t = TypeWithProperty()
t.X(1) // this will "call" property
```

vzarytovskii: I presume it's not a breaking change in this sense, could you please make an explicit comment, that it's not?

vzarytovskii:
I'm fine to merge it with clarifications that it's not going to shadow delegate properties, and so on.

PR can be merged independently.

preview/FS-1137-prefer-extensions-method-over-property-when-arguments-provided.md
smoothdeveloper: It still calls the property due to check on property type being `TType_fun` to ignore looking up extension methods.

https://github.com/dotnet/fsharp/blob/6a9791f6cc5f6e59ca729917e991b827d00541dc/src/Compiler/Checking/NameResolution.fs#L2644-L2646

I tried declaring events, they don't go to the same "Property" branch, and regarding properties that would be of a delegate type, in F# they are always invoked with `.Invoke` so it shortcuts as having no "delayed items" for which type checking works the same as before.

Based on how it is specified (things that can't be shadowed) and how it is implemented, I'm overall confident that it won't be a breaking change to existing code.

The edge case that would be breaking I thought about is indexed properties overloads for a non indexed property but F# doesn't allow to specify this:
```fsharp
type C() =
    member _.Y = 1
    member _.Y with get (v,v2) = printfn $"Y({v},{v2})"
```

> error FS0436: The property 'Y' has the same name as another property in type 'C', but one takes indexer arguments and the other does not. You may be missing an indexer argument to one of your properties.

smoothdeveloper:
This can be merged.

PR #764, opened 2024-02-10
add [fs-1141] Allow access modifiers to auto properties getters and setters

Tangent-90:

### DIFF
--- a/RFCs/FS-1141-Allow-access-modifiers-to-auto-properties-getters-and-setters.md
+++ b/RFCs/FS-1141-Allow-access-modifiers-to-auto-properties-getters-and-setters.md
@@ -20,9 +20,9 @@ To simplify the syntax of defining a private-settable-read-only properties.
 
 # Detailed design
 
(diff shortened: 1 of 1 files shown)


RFCs/FS-1141-Allow-access-modifiers-to-auto-properties-getters-and-setters.md
```diff
@@ -0,0 +1,124 @@
...
+
+# Summary
+
+Allow access modifiers to auto properties getters and setters.
+
+# Motivation
+
+To simplify the syntax of defining a private-settable-read-only properties.
+
+# Detailed design
```
cartermp, line 21: I don't think this is detailed enough. What does the new syntax compile down into? Which access modifiers are allowed? What can be mixed or not mixed? etc.

Tangent-90: updated

PR #767, opened 2024-04-16
FS-1124: Clarify static abstract methods called on interfaces

edgarfgp:

### DIFF
--- a/FSharp-7.0/FS-1124-interfaces-with-static-abstract-members.md
+++ b/FSharp-7.0/FS-1124-interfaces-with-static-abstract-members.md
@@ -57,6 +57,8 @@ type C() =
 
 Static abstract methods may not be declared in classes.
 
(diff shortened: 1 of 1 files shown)


FSharp-7.0/FS-1124-interfaces-with-static-abstract-members.md
```diff
@@ -57,6 +57,10 @@ type C() =
 
 Static abstract methods may not be declared in classes.
 
+Static abstract methods may not be called directly on interfaces.
```
vzarytovskii, line 60: It should be explicitly mentioned that virtual methods can be called directly on the interfaces.

edgarfgp: Done. Thanks for the review

vzarytovskii (approved):
Thanks!

PR #771, opened 2024-06-04
[FS-1143] Generic Attributes

mflibby:


dsyme:
@T-Gro @vzarytovskii I'd be happy to accept this RFC. Could you assess and if we agree then merge? thanks

T-Gro (approved):

T-Gro:
This RFC is good.
There are remaining implementation challenges which are being discussed in the PR, most notably about pickling the information and impact it has on backwards compat (especially not breaking it for code which does not use this feature).

PR #776, opened 2024-06-28
[RFC FS-1145] C# collection expression support for lists & sets

brianrourkeboll:
RFC for approved language suggestion https://github.com/fsharp/fslang-suggestions/issues/1355.

Discussion thread: TODO
Proposed implementation: https://github.com/dotnet/fsharp/pull/17359

### DIFF
--- a/RFCs/FS-1145-csharp-collection-expression-support-for-lists-and-sets.md
+++ b/RFCs/FS-1145-csharp-collection-expression-support-for-lists-and-sets.md
@@ -1,0 +1,331 @@
+# F# RFC FS-1145 - C# collection expression support for F# lists and sets
+
+The design suggestion [Support for C# collection expressions in F# lists and sets](https://github.com/fsharp/fslang-suggestions/issues/1355) has been marked "approved in principle."
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/1355)
+- [x] Approved in principle
+- [x] [Implementation](https://github.com/dotnet/fsharp/pull/17359)
+- [ ] [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)
+
+# Summary
+
+<!-- One paragraph explanation of the feature. -->
+
+We annotate the `'T list` and `Set<'T>` types in FSharp.Core with `CollectionBuilderAttribute` and expose static `Create` methods to enable instantiating F# lists and sets in C# using C# 12's [collection expressions](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-12.0/collection-expressions) feature.
+
+## Before
+
+### Instantiation of intermediate collections required
+
+```csharp
+FSharpList<int> xs = ListModule.OfArray(new[] { 1, 2, 3 });
+```
+
+```csharp
+FSharpSet<int> xs = SetModule.OfArray(new[] { 1, 2, 3 });
+```
+
+## After
+
+### Instantiation of intermediate collections no longer required
+
+```csharp
+FSharpList<int> xs = [1, 2, 3];
+```
+
+```csharp
+FSharpSet<int> xs = [1, 2, 3];
+```
+
+# Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected outcome? -->
+
+Just as it is common for F# code to consume .NET libraries written in C#, there is a variety of scenarios where it is necessary to interact with F# libraries from C#.
+
+In mixed-language solutions containing both C# and F# projects, for example, unit tests written in C# may need to test APIs written in F#.
+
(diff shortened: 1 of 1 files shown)


RFCs/FS-1145-csharp-collection-expression-support-for-lists-and-sets.md
```diff
@@ -232,7 +236,7 @@ type Set<'T> = …
 
 <!-- Why should we *not* do this? -->
 
-There are no obvious drawbacks to adding this support to FSharp.Core.
+The main arguable drawback of this approach is that the API surface area of FSharp.Core is now technically different depending on the target framework. In practice, we mitigate this by annotating the new APIs with the `CompilerMessageAttribute` and specifying `IsHidden=true`. This means that editors will not suggest `List.Create` and `Set.Create` in autocompletion, and, if a user nonetheless attempts to use them directly, the compiler will emit a warning that these APIs are intended for compiler use only.
```
baronfel, line 239: Have you considered the use of the [EditorBrowsableAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.editorbrowsableattribute?view=net-8.0) instead? That's the one I'm most familiar with editors and tooling using as a 'hide me' marker.

brianrourkeboll: Yes, that's a possibility. I merely went with `CompilerMessageAttribute` because that's the attribute used for most other constructs defined in FSharp.Core that aren't meant to be used directly outside of it. We could apply both attributes here, although VS and Ionide do already respect `CompilerMessageAttribute` on its own (maybe something in FCS completion is aware of it).

brianrourkeboll: I think we would at least want to keep `CompilerMessageAttribute` for the warning (FS1204) either way.

brianrourkeboll: I think I'm just going to leave it like this for now.

Only `CompilerMessageAttribute` is used in FSharp.Core at present:

https://github.com/search?q=repo%3Adotnet%2Ffsharp+%5B%3CCompilerMessage+language%3AF%23&type=code

`EditorBrowsableAttribute` is not:

https://github.com/search?q=repo%3Adotnet%2Ffsharp+%5B%3CEditorBrowsable+language%3AF%23&type=code

The name completion logic in FCS accounts for the presence of `CompilerMessageAttribute` with `IsHidden=true` here:

- https://github.com/dotnet/fsharp/blob/df3382ce9d0d66b7e2e3830af5826b0f2dd49289/src/Compiler/Checking/NameResolution.fs#L4336
- https://github.com/dotnet/fsharp/blob/df3382ce9d0d66b7e2e3830af5826b0f2dd49289/src/Compiler/Checking/NameResolution.fs#L4448
- https://github.com/dotnet/fsharp/blob/df3382ce9d0d66b7e2e3830af5826b0f2dd49289/src/Compiler/Checking/AttributeChecking.fs#L452
- https://github.com/dotnet/fsharp/blob/df3382ce9d0d66b7e2e3830af5826b0f2dd49289/src/Compiler/Checking/AttributeChecking.fs#L375
- https://github.com/dotnet/fsharp/blob/df3382ce9d0d66b7e2e3830af5826b0f2dd49289/src/Compiler/Checking/AttributeChecking.fs#L353-L355

If we wanted to, down the line, we could make a separate update to FSharp.Core to add `EditorBrowsableAttribute` to every construct that currently has `CompilerMessageAttribute` with `IsHidden=true`.

baronfel: That's very reasonable - just wanted to raise this as a discussion point because of where I'd seen EditorBrowseable used in other designs on dotnet/designs and dotnet/csharplang.

brianrourkeboll: Yeah, and `EditorBrowsableAttribute` would make sense if we also wanted to hide these types and methods from completion in C#, although I'm not sure that's necessary for this feature.

brianrourkeboll: And I very much appreciate your taking the time to look and raise the question!

brianrourkeboll: > Yeah, and `EditorBrowsableAttribute` would make sense if we also wanted to hide these types and methods from completion in C#, although I'm not sure that's necessary for this feature. The primary reason to hide them in F# is, I think, to maintain parity of the apparent API surface area of FSharp.Core across target frameworks. Do you see a reason to hide them from C#?

For what it's worth, the `Create` methods like this in the BCL are _not_ hidden.

psfinaki:
@vzarytovskii probably should be merged, the PR is about to get in.

PR #783, opened 2024-08-08
RFC-1147 for non-string args for compiler directives

Martin521:
While working on the "scoped nowarn" RFC, I wanted to reference this one but found it didn't exist yet.


vzarytovskii (approved):

vzarytovskii:
Thanks!

PR #785, opened 2024-08-16
Allow object expression without overrides.

edgarfgp:


vzarytovskii:
From what I understand, there's a difference from the original suggesion, which is: we allow any member-less expressions in object expressions, an not only for abstract classes, but regular ones as well (as you mentioned here: https://github.com/fsharp/fslang-suggestions/issues/632#issuecomment-2291060478), I'm fine with it as long as @dsyme is, as it feels more uniformed, despite the fact it was explicitly disallowed in the past.

dsyme (approved):

dsyme:
This looks great to me!

PR #788, opened 2024-09-22
RFC FS-1033 - Deprecate places where `seq` can be omitted.

edgarfgp:


RFCs/FS-1033-Deprecate-places-where-seq-can-be-omitted.md
```diff
@@ -0,0 +1,313 @@
...
+
+- Having only one way to create a sequence will make the language more consistent and easier to understand.
+
+- Prevents the accidental override of index range operator.
+```fsharp
+let (..) _ _ = "Im an index range operator"
+
+let x = { 1..10 } // Here the index range operator is being overridden and `x` will be a string instead of a sequence.
+```
+
```
Martin521, line 34: ``suggestion
- The FSharp-Spec-4.1 already mentions that the ommission of the `seq` keyword is not recommended and that it might be deprecated (§6.3.12).

```

File: RFCs/FS-1033-Deprecate-places-where-seq-can-be-omitted.md [hunk 2]
```diff
@@ -0,0 +1,313 @@
...
+- Having only one way to create a sequence will make the language more consistent and easier to understand.
+
+- Prevents the accidental override of index range operator.
+```fsharp
+let (..) _ _ = "Im an index range operator"
+
+let x = { 1..10 } // Here the index range operator is being overridden and `x` will be a string instead of a sequence.
+```
```
Martin521, line 33: Will that be different when using `seq {1..10}`? According to the spec, also in this case the `(..)` operator will be used.

Martin521 (approved):

RFCs/FS-1033-Deprecate-places-where-seq-can-be-omitted.md
edgarfgp: - With index range operator being overridden

```fsharp
let (..) _ _ = "Im an index range operator"

{ 1..10 } // inferred as string

let str = { 1..10 } // inferred as string

let charSeq = seq  { 1..10 } // inferred  as char seq
```

- Without index range operator being overridden

```fsharp
{ 1..10 } // inferred as int seq

let str = { 1..10 } // inferred as int seq

let intSeq = seq  { 1..10 } // inferred  as int seq
```

Martin521: Interesting. The [spec says](https://github.com/fsharp/fslang-spec/blob/main/releases/FSharp-Spec-4.1.2024-10-02.md#6312-range-expressions)

Range expressions involving `expr1 .. expr2` are translated to uses of the `(..)` operator, and those
involving `expr1 .. expr1 .. expr3` are translated to uses of the `(.. ..)` operator:

```fsgrammar
seq { e1 .. e2 } → ( .. ) e1 e2
seq { e1 .. e2 .. e3 } → ( .. .. ) e1 e2 e3
```

So this is separate issue that requires either fixing the compiler or fixing the spec.

edgarfgp: > Interesting. The [spec says](https://github.com/fsharp/fslang-spec/blob/main/releases/FSharp-Spec-4.1.2024-10-02.md#6312-range-expressions)
> 
> Range expressions involving `expr1 .. expr2` are translated to uses of the `(..)` operator, and those involving `expr1 .. expr1 .. expr3` are translated to uses of the `(.. ..)` operator:
> 
> ```
> seq { e1 .. e2 } → ( .. ) e1 e2
> seq { e1 .. e2 .. e3 } → ( .. .. ) e1 e2 e3
> ```
> 
> So this is separate issue that requires either fixing the compiler or fixing the spec.

Looking at the [compiler](https://github.com/edgarfgp/fsharp/blob/c68a3012a91217209452a949f866241d2d7bba32/src/Compiler/Checking/Expressions/CheckExpressionsOps.fs#L151C1-L154C89) it is as per spec describes. 

The main issue is that you can overload the  `..` op_Range and `.. ..`   op_RangeStep and when using these you are not longer creating a sequence.

edgarfgp:
This is ready.

RFCs/FS-1033-Deprecate-places-where-seq-can-be-omitted.md
Martin521: Ok, it took me some time, but now I think I know what's happening.

1. Overriding the `(..)` is and should be allowed, like for all other operators. 
    (Like for all operators, global override is not recommended).
3. The overridden `(..)` operator is used in evaluating the range expression `{1..10}`.
4. If you override `(..)` to return a string, then `{1..10}` returns a string. 
5. If you now apply `seq` to that string, you get a `seq<char>`.

All of this is ok and should stay as is. We just want to deprecate the standalone use of range expressions.
I therefore recommend to remove the bullet item "Preventing the overload ..."

(If we really would want to prohibit overriding `(..)`, we could go a step further and change the nature of `..` from being an operator to being a special syntax element that is allowed only in comprehensions and sequence iteration expressions. But this is probably a risky step in terms of compatibility, and also does not feel FSharp-ish.)

PR #794, opened 2025-01-31
Update F# 9 RFCs

jwosty:
Update all F# 9 RFCs (AFAIK, please verify) - move to new F# 9 folders, fix up links, and fix up some formatting.


FSharp-9.0/FS-1043-extension-members-for-operators-and-srtp-constraints.md
```diff
@@ -8,8 +8,8 @@ These design suggestions:
 have been marked "approved in principle". This RFC covers the detailed proposal for these
 
 * [x] Approved in principle
-* [Discussion thread](https://github.com/fsharp/fslang-design/issues/435)
-* [ ] Implementation: [In progress](https://github.com/dotnet/fsharp/pull/6805)
+* [x] [Discussion](https://github.com/fsharp/fslang-design/issues/435)
+* [x] [Implementation](https://github.com/dotnet/fsharp/pull/8404)
```
vzarytovskii, line 12: This one isn't in F# 9

vzarytovskii (changes requested):
One isn't in F# 9, rest seems to be ok, thanks

FSharp-9.0/FS-1043-extension-members-for-operators-and-srtp-constraints.md
jwosty: Ah, I see my mixup. Was there an RFC for the "Prefer extension methods to intrinsic properties when arguments are provided" bit that was added in F# 9? https://learn.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-9#prefer-extension-methods-to-intrinsic-properties-when-arguments-are-provided

Martin521:
[This one](https://github.com/fsharp/fslang-design/blob/main/drafts/FS-1147-non-string-directive-args.md) was also implemented and should be moved.

Martin521:
I have a few remarks on FS-1060. Not sure if they fit here or should be filed as an issue for that RFC.

I am not sure if [`#nowarn "nullness"`](https://github.com/fsharp/fslang-design/blob/e0db7c6e278731489466394a94aabe820cf5d68f/FSharp-9.0/FS-1060-nullable-reference-types.md?plain=1#L769) was not implemented.

The section on [project configuration](https://github.com/fsharp/fslang-design/blob/e0db7c6e278731489466394a94aabe820cf5d68f/FSharp-9.0/FS-1060-nullable-reference-types.md?plain=1#L777) contains initial thoughts on requirements, but not the concrete spec of the final implementation. You could argue that this whole section is not about the language and does not need to be in the RFC. But then some of the items (like `--checknulls` and `<Nullable>enable</Nullable>`) appear in the language guide (which should be based on the spec). And I don't know if the other items are implemented or not.

And the link target in [this](https://github.com/fsharp/fslang-design/blob/e0db7c6e278731489466394a94aabe820cf5d68f/FSharp-9.0/FS-1060-nullable-reference-types.md?plain=1#L779) line seems not to exist.

Martin521:
FS-1079 should mention that the feature does not apply to single-case unions.

FSharp-9.0/FS-1043-extension-members-for-operators-and-srtp-constraints.md
jwosty: Moved FS-1043 back

jwosty:
> [This one](https://github.com/fsharp/fslang-design/blob/main/drafts/FS-1147-non-string-directive-args.md) was also implemented and should be moved to the new folder.

Good catch. Moved.

> I have a few remarks on FS-1060. Not sure if they fit here or should be filed as an issue for that RFC.

I would file a new discussion for that.

T-Gro:
> I have a few remarks on FS-1060. Not sure if they fit here or should be filed as an issue for that RFC.

I would recommend to finalize this "movement" PR from @jwosty and deal with individual RFC contents separately (discussion or a dedicated PR suggesting changes).

As an example for FS-1060, the whole "project config in VS" section was probably good material for facilitating discussion at time of origin, but does not need to keep living in the spec => It can be either removed, or remain in part of the RFC that is not yet finished (i.e. not moved to a langversion-correspondant folder) and is open up for implementation.

jwosty:
@T-Gro @vzarytovskii can we get this merged? Until this gets merged there's no "official" list of PR's that made it into F# 9 :)

PR #795, opened 2025-02-02
Add that FS-1079 is only for multi-case DUs

Martin521:
See also [this](https://github.com/fsharp/fslang-suggestions/issues/1394) discussion.

Note: this should be merged before moving the RFC.

### DIFF
--- a/RFCs/FS-1079-union-properties-visible.md
+++ b/RFCs/FS-1079-union-properties-visible.md
@@ -12,7 +12,7 @@ This RFC covers the detailed proposal.
 
 F# discriminated unions have a number of generated public members that were created for C# interoperability and are hidden from F# code. These include an instance property `IsFoo : bool` per case `Foo`.
 
-This proposal is to make these members visible and usable from F# code too.
+This proposal is to make, for discriminated unions with more than one case, these members visible and usable from F# code too.
 
 # Motivation
 


vzarytovskii (approved):

charlesroddie:
This PR rewrites the history and makes the document internally inconsistent. "This proposal is to make, for discriminated unions with more than one case" vs "The compiler generates the following public members for every discriminated union type:". The original RFC was just incompletely implemented. Restricting FS-1079 is a new (and bad) idea.

Martin521:
It is quite normal that a RFC is still changed during implementation and that has nothing to do with rewriting history - the history is maintained in git.
I don't want to judge here if the change was a good or a bad one. That's a different discussion.
But I do want a true spec of F# 9, and FS-1079 is now a F# 9 RFC.

Of course, theoretically, if F# had a huge team with standardization committees etc, the spec would be in the lead and the implementation had to follow. But de facto the one and only F# compiler is the reference.

It is bad enough that many RFCs are - even though approved - in a bad shape. But I want at least to correct the most obvious discrepancies between implementation and spec.

If F# 10 gets single-case public `Is*`, that needs a new RFC.

I am sorry I created an inconsistency by missing the second occurrence of the sentence. I could create another PR for that, but I also don't want to bother the team with too many nitpicks.

T-Gro:
This text change is after all a cosmetic change. The previous text had that behavior implicit in the sentence `F# discriminated unions have a number of generated public members that were created for C# interoperability` - by leaning onto the pre-existing codegen situation. But of course it was a low level detail which might not have been immediately apparent.

The newly added wording makes that clearer, and reflects what is part of F# 9.

PR #800, opened 2025-05-30
Create FS-1150-Literal-inferred-types.md

Happypig375:


Happypig375:
pending changes from https://github.com/fsharp/fslang-suggestions/issues/1427#issuecomment-2935564324

T-Gro (commented):
I think it would be good to emphasize that the biggest advantage is not with explicit type annotation for standalone values (seen here in examples but also typical for C# not using `var`).

The motivating examples should come from passing arguments to existing methods/functions. The benefits are then multi-fold:
- Easier code with fewer boilerplate
- Fewer runtime conversions
- Possibilities for "improve perf by recompiling" scenarios, e.g. if a library changes it's arguments to support stack allocated ROS.

T-Gro (commented):
One aspect I would like to see in bigger detail, e.g. as a diff to existing spec, is how this plays with method resolution in case of multiple ambiguous overloads - non-generic as well as generic. Since this RFC touches literals, it does not have much overlap with function parameter inference - even though scenarios combining literals and symbols like 
```fsharp

let myFunc x =
    let _ = [ 1; x ; 3]
```
mean it has to be considered as well.

T-Gro:
> @T-Gro Should there be new static constraints associated with these inferences?

Could you elaborate please? I am not sure I follow what exactly you mean.

Happypig375:
@T-Gro 

```fs
let inline f a b = a + b
let a: int64 = f 1L 2L
let b: byte = f 1uy 2uy
// works similarly to
let inline g() = 1
let c: int64 = g()
let d: byte = g()
```
then this inline function needs a new static constraint, yes?

This also means a potential breaking change
```fs
let f a b = a + b
let a = f 1uy 2uy // f: byte -> byte -> byte
let b = f 1 2 // error
// similarly...
let x = 1
let y: int64 = x // infers x as int64
let z: int32 = x // error?
```

T-Gro:
> let inline myFunc x = [1; x; 3]
> // val inline myFunc: ^a -> ^b when ^a: 1 and ^a: 3 and ^b: [^a]

So the type constraint would be with the semantics ` 'T when 'T can be used to express literal "1"` ?

I haven't thought about such deep and invasive impact on type inference - this would likely change any piece of code working with literals in basic arithmetic ways. This probably mean that the feature would have to be rejected due to the impact on existing code.

Happypig375:
@T-Gro But it would also enable more beginner friendliness like
```fs
1 + 1.5 + 2
```

such invasive impacts on type inference is restricted to `inline` code only. For any code that wishes to fix types at e.g. API boundaries, just use type annotations.

Happypig375:
```fs
open System.Numerics
type Vector3 with
    static member op_Implicit struct(x, y, z) = Vector3(x, y, z)
let x: Vector3 list = [
    1, 2, 3
    4, 5, 6
]
```

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,523 @@
...
+    "alice", 23
+    "bob", 34
+    "carol", 55
+]
+```
+
+List literals will include `with` as a computation expression keyword. It must be placed at the start of the list literal. Its use will be the same as in C#'s collection expressions. The syntax to the right of `with` is the constructor call syntax, this means that it can either take an argument application without parentheses or tupled arguments with optionally named parameters.
+```fs
+let l: ResizeArray<int> = [with(capacity = 3); 1; 2]
+```
```
brianrourkeboll, line 436: Hmm, I understand that part of the point of adding type-direction to `[…]` is to make it "the one collection expression syntax to rule them all," including in patterns, where curly-brace pattern syntax that is symmetrical with the construction syntax, like `seq { … }`, is not currently valid. From that perspective, I understand the desire to add a way to specify constructor arguments to `[…]` so that it can effectively subsume _all_ existing ways of constructing collections.

But this `[with …; …]` syntax in particular feels pretty weird to me in C#, and I think it would feel even weirder in F#.

It is already possible, for example, to make variants of computation expressions like those in https://github.com/fsprojects/FSharp.Collections.Builders that take initial capacities or comparers:

```fsharp
let nameToAge2 =
    dictionary StringComparer.OrdinalIgnoreCase {
        "alice", 23
        "bob", 34
        "carol", 55
    }
```

```fsharp
let l = resizeArray 3 { 1; 2 }
```

As for making a computation-expression-builder-like syntax of that kind work in a backwards-compatible way with type-directed `[…]`: while it might be nice, it seems like it might be difficult, since things like this are already valid:

```fsharp
let ResizeArray (capacity : int) (collection : _ seq) =
    let xs = ResizeArray capacity
    xs.AddRange collection
    xs

let xs = ResizeArray 3 [1; 2]
```

But since you already _can_ define computation expressions that achieve the same result with syntax that is already valid and, I think, better anyway (the curly braces notwithstanding), I am not convinced of the need to add new syntax as odd and special as `[with …; …]`.

It seems like this syntax could also present problems when combined with type inference.

For example, if type-directed `[…]` made this valid:

```fsharp
let xs = [1; 2]
let ys : int array = xs
```

Would this also be valid?

```fsharp
let xs = [with(capacity = 3); 1; 2]
let ys : ResizeArray<int> = xs
```

If so, what would the compiler error look like here?

```fsharp
let xs = [with(capacity = 3); 1; 2]
let ys : int array = xs
```

Would something like this be valid?

```fsharp
let inline f capacity = [with(capacity); 1; 2]
let xs : ResizeArray<int> = f 3
```

Or this?

```fsharp
let inline f capacity comparer = [with(capacity, comparer); "a", 1; "b", 2]
let xs : Dictionary<string, int> = f 3 StringComparer.Ordinal
```

That kind of thing seems like it would necessitate adding a new way to specify constraints for arbitrary constructor signatures.

Happypig375: Hmm you're right. Since constructor arguments are intrinsically tied to the type to be inferred, it might instead make sense to do something like
```fs
ResizeArray<int> 3 |> init [1; 2; 3]
```
?

brianrourkeboll: > Hmm you're right. Since constructor arguments are intrinsically tied to the type to be inferred, it might instead make sense to do something like
> 
> ```fsharp
> ResizeArray<int> 3 |> init [1; 2; 3]
> ```
> 
> ?

Hmm, but that syntax is of course already valid. Somebody could already have something like

```fsharp
let init (xs : _ seq) (resizeArray : ResizeArray<_>) =
    resizeArray.AddRange xs
    resizeArray
```

and the example above would compile.

...What if we were a bit creative here and looked at keywords...

Since curried constructors are not allowed, I wonder if we could do something like this where we required the `new` keyword:

```fsharp
let xs = new ResizeArray<int> (capacity=3) [1; 2]
```

The above is _not_ currently valid, which means that we could make it mean what we would want it to mean here.

On the other hand, if I recall correctly, the reason C# did _not_ go with something like this is because the C# collection expression design also involves calling static collection builder methods or using special, compiler-generated code to initialize the collection, not just constructors.

If we wanted to do the same (which would be a reasonable thing to do, since it would let us likewise take advantage of the collection builder methods that have been added to the BCL), then I guess we might also not want to use `new` for this.

Happypig375: `init` could be a library function to be interpreted specially by the compiler when used.

Happypig375:
@T-Gro There are 2 ways to think about the definition of integer (base 10) and float literals without separators:

A) Float literals are expressed as mantissa * 10^exponent. Integer (base 10) literals are a subset of float literals without non-zero digits after the unit digit.
```fs
let x: int = 1.000 // allowed
let y: bigint = 2.66e200 // allowed
let z: int = 2.67e1 // error: non-zero digits after unit place
```

B) Integer (base 10) literals are groups of 0-9. Float literals are a superset of integer (base 10) literals that allow the decimal dot or exponent.
```fs
let x: int = 1.000 // error: decimal dot. Write as 1 instead.
let y: bigint = 2.66e200 // error: decimal dot, exponent. Write as 266 * 10 ** 198 instead.
let z: int = 2.67e1 // error: decimal dot, exponent 
```

It seems that A) is more preferable for flexibility in specification of integers.

As for the inherent imprecision of `0.3` in binary floating point numbers, since `0.3: float` is already allowed today, I don't think it should be disallowed. Instead, whether an integer or float literal can fit inside a numeric type can follow the general guidelines already present in the docs and many learning materials:

Type | Constraints
-|-
`int8` `uint8` `int16` `uint16` `int32` `uint32` `int64` `uint64` | `MinValue` < n < `MaxValue`
`nativeint` `unativeint` | `MinValue` of 32-bit type < n < `MaxValue` or 32-bit type
`float32` `float` `decimal` | 1) `MinValue` < n < `MaxValue` 2) `Precision` (7 for `float32`, 15 for `float`, 28 for `decimal`) <= #digits

For floating point numbers, the "closest expressible value" will continue to be the result for float literals that cannot be expressed precisely.

RFCs/FS-1150-Literal-inferred-types.md
Happypig375: C# has decided that `with` will not participate in overload resolution. Therefore, checking of `with` is done after type inference; error on the `with`.

Happypig375: I rewrote that part to not use `with` anymore.

Happypig375:
It's ready for second round review

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,749 @@
...
+// val inline e: unit -> ^a when ^a: 10000
+let inline f() = 1.00001e+4
+// val inline f: unit -> ^a when ^a: 10000.1 and ^a: float
+let inline g() = 1.0000e+4
+// val inline g: unit -> ^a when ^a: 10000
+let inline h() = 1.00000e+4 // Note the final 0 falls behind the unit place
+// val inline h: unit -> ^a when ^a: 10000 and ^a: float
+```
+
+The numeric value constraint can be satisfied by the following types, in order of method overload preference, applying if the numeric value is in range:
```
T-Gro, line 238: This needs more detail w.r.t. to methods with multiple arguments.

```
M(int,float,bigint)
M(bigint,bigint,bigint)
M(float,bigint,int)

M(1,1,1)
let a = 2
M(a,a,a)

```

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,749 @@
...
+
+```fs
+module A
+let a = 1 // This is public. There may be external dependencies.
+let b = a + 1L
+// before: Uses type-directed conversion of int32 -> int64.
+// after: changes type of a to int64.
+```
+
+This can easily be solved with a recompilation of consuming code.
```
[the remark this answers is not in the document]
Happypig375: Hmm, then the existing statically resolved type constraint solver would need to be updated for constraints that cannot persist across `public` boundaries. Is this what this implies?

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,749 @@
...
+
+F# char literals `'c'` would not just be resolved to `char`, but also `byte` and so on.
+
+F# tuple literals would not just be resolved to Tuple (`'a * 'b`), but also struct tuple (`struct('a * 'b)`) and `KeyValuePair<_, _>`. There seems to be not much value to extend this syntax to other types.
+
+F# list literals `[]` would not just be resolved to `list`, but also `ImmutableArray<_>`, `ReadOnlySpan<_>` and so on.
+
+F# string literals `"abc"` would not just be resolved to `string`, but also `PrintfFormat`, `char array`, `ReadOnlySpan<byte>`, `Rune list` and so on.
+
+F# option literals (and other DU values) `Some` would not just be resolved to `option` but also `voption` if an alias is defined. Similarly for the union case `Width` which can refer to the discriminated union types `HTMLAttr`, `CSSProp` or `SVGAttr` in `Fable.React.Props`.
```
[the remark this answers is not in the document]
Happypig375: Extracted to https://github.com/fsharp/fslang-suggestions/issues/1074#issuecomment-2995866188

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,627 @@
...
+- Implicit `op_Implicit` conversions help more than it hinders e.g. going from some concrete type to a base class, and it's pretty easy to explain. But, it'll require some explanation of what `op_Implicit` is etc. - a completely foreign concept for everyday F# developers.
+- Implicit type-directed resolutions of literals require explanation of constructors and builder patterns, and the fact that `let x : Set = [ 1 .. 10 ]` isn't the same from a performance point of view as `let x = Set [ 1 .. 10 ]` will be challenging.
+
+## Diagnostics
+There is also risk of introducing action-at-a-distance type resolution behaviour when editing F# code.
+```fs
+let a = 1 // Defaults to int
+let b = 2 // Without code below, this defaults to int.
+...
+let c = b + 1.5 // Without code below, this defaults b and c to float instead of int,
```
Martin521, line 98: Not true. This is an error in F#.

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,627 @@
...
+...
+let d: float32 = c // This fixes c, and therefore b, to float32.
+```
+This can be mitigated with two potential approaches:
+1. a warning that type defaulting behaviour is used. For example, `let x = 1` without further type restriction.
+2. exposing the behaviour of type defaulting first-class. This is covered in a separate suggestion - [Display type defaulting](https://github.com/fsharp/fslang-suggestions/issues/1427).
+
```
Martin521, line 105: The Diagnostics section should include an opt-in warning that can be enabled to warn when the new features are used.

RFCs/FS-1150-Literal-inferred-types.md
```diff
@@ -0,0 +1,627 @@
...
+## Breaking changes
+
+There are potential breaking changes with interactions on previously defined type-directed conversions. Specifically, around `int64`, `nativeint` and `float` as defined in [FS-1093](https://github.com/fsharp/fslang-design/blob/main/FSharp-6.0/FS-1093-additional-conversions.md).
+
+```fs
+module A
+let a = 1 // This is public. There may be external dependencies.
+let b = a + 1L
+// before: Uses type-directed conversion of int32 -> int64.
```
Martin521, line 114: Not true. The previous line is an error in F#.

Martin521 (commented):
Here are two general comments.
a) I believe the main purpose of an RFC is to specify the new features in detail. This should include in detail the changes that would be necessary in the F# spec. I know this hasn't been done in many RFCs, but I think is is important if we want to call F# a language and not just a compiler. 
b) The sections on motivation and drawbacks are very one-sided. This should be avoided in RFCs.

RFCs/FS-1150-Literal-inferred-types.md
Happypig375: Hmm, it is indeed missing the explicit type annotation.

Happypig375: Fixed the sample

Happypig375:
@Martin521 What is your suggestion on avoiding one sidedness then?

T-Gro:
> Here are two general comments. a) I believe the main purpose of an RFC is to specify the new features in detail. This should include in detail the changes that would be necessary in the F# spec. I know this hasn't been done in many RFCs, but I think is is important if we want to call F# a language and not just a compiler. b) The sections on motivation and drawbacks are very one-sided. This should be avoided in RFCs.

I agree especially when it comes to name resolution and type inference.
The algorithm/decision tree for propper resolution is quite complex already.

This RFC would introduce:
- New ruleset based on ordered prefence lists for each literal kind
- Further changes to prioritization of overloads
- A "tie breaker" rule to avoid amiguity errors

The RFC should indicate how the spec will change, and aim to use existing terminology from the spec.

Happypig375:
I realized that this would also change
```fs
let a = ["a"; "b"; "c"]
let b = a :> obj // Should this now produce an error after inferring as ReadOnlySpan?
System.String.Concat(",", a)
```

T-Gro:
> I realized that this would also change
> 
> ```fsharp
> let a = ["a"; "b"; "c"]
> let b = a :> obj // Should this now produce an error after inferring as ReadOnlySpan?
> System.String.Concat(",", a)
> ```

In the general case, this is hard - type inference works top to bottom, and doing a decision system which can effected previously decided bindings will be a massive change.

In this specific case, where a warning comes from PostInferenceChecks (post-inference), like the case with byref safety, this could likely be achieved like you write - infer based on `Concat` call, and checks would then complain about the `:> obj` cast.

This however introduces inconsistency, or rather leaks the implementation details of checks (during inference or after it) to the end user.

T-Gro:
@Happypig375 :

I will do another review round during August and will come back to the suggestion with my current opinions ( approve in principle | generally agree | undecided | probably not) on each sub proposal listed in the RFC.


Now that we have the details written out (thanks 👍  again for doing that, I think it does bring clarity into the breadth of the proposal and the different phases that can be decided and implemented separately), we can refer to them and achieve a more focused discussion.

Happypig375:
@T-Gro 👍 My attention is currently diverted to fold loops so this can be on hold for now.

Happypig375:
Are only non-draft PRs reviewed?

T-Gro:
I have reverted some other open RFCs back to draft status so that I do not need to revisit them again - it is an indicator that it awaits either further discussion, reaction to feedback, or some final touches.

For this PR in particular, I do not want to rush the decision - many of the subitems carry significant change with them, and I particularly careful with hard-to-revert decisions. Compiler perf for now-common-idioms is an aspect of it, especially for new mechanics using SRTP-like resolution under the cover.


I apologize for taking a longer time and I know I wrote I will do that sooner.

T-Gro:
Once again, thank you @Happypig375  for the work on this RFC. 
The level of detail and consideration for edge cases and gradual improvement of the feature trough specified layers is evident

After considering the proposal in depth, here is the recommendation:

---

### Approved separately

The following sections are approved and should proceed as separate, focused RFCs:

| Section | Feature | Suggestion link |
|---------|---------|--------|
| **a (Levels 1-2 only)** | Type-directed literals at method arguments and explicit annotations |   Part of https://github.com/fsharp/fslang-suggestions/issues/1421 |
| **e** | float64 type abbreviation and d literal suffix | New (consistency with int32/int64) |
| **h** | Deconstruct pattern support | Was approved https://github.com/fsharp/fslang-suggestions/issues/751 |
| **r** | UTF-8 B-suffix string literals | Part of https://github.com/fsharp/fslang-suggestions/issues/1421 |
| **t** | Record update syntax for C# records | Approving in principle now https://github.com/fsharp/fslang-suggestions/issues/1138 |

---

### Not Approved for F# that exists today

The following 3 categories would require significant changes to type inference and the constraint system, and right now I do not see the direct benefits outweigh the cost coming with added complexity, potential bugs and tooling+ecosystem changes.
F# exists in an ecosystem with variety of tools and libraries and full set of proposals would change how components interact for pretty much any non-trivial code.

I do like having them written out and described in detail, and they might serve very well for an experimental greenfield prototype demonstrating another possibility for F# or its descendants. Such prototype/language would need to make different assumptions about SRTP constraints and put them into the very core of the language and the ecosystem. 

If any student looking for thesis material reads this, such prototype on bringing duck typing in a statically type .NET language could serve very well 👍 . 

**Category 1: Backwards Type Inference for Literals (a3-a7)**

F# already supports deferred type resolution for generic constructs like `let add x y = x + y`. However, literals like `1` and `[1;2;3]` have historically resolved to fixed default types (`int`, `list`). Extending backwards inference to literals introduces a new category of action-at-distance where the meaning of simple expressions changes based on code added later. This is a larger shift in expected behavior than existing generic inference patterns.

**Category 2: New SRTP Constraint Forms (b, f, g, k, p, q)**

The proposed numeric value constraints (`^a: 1`), range constraints (`^a: 1..10`), tuple constraints (`^a: (^b, ^c)`), list constraints (`^a: [^b]`), and measurable constraints represent substantial additions to the constraint system. The implementation cost and impact on error messages, tooling, and developer experience outweigh the syntactic convenience.

**Category 3 Semantic Redefinition (c, q)**

Type-directed numeric literals (c) and string literals (q) would change the meaning of `1` and `"abc"` based on downstream context. Unlike generic functions where polymorphism is expected, literals have traditionally been concrete. Changing this creates debugging and comprehension challenges that outweigh the benefits.


### Need more time / deferred

**Deferred Pending Dependencies (d, l)**

Section d requires SRTP field constraints which are not yet implemented. Section l requires FS-1023 (type providers generating from types). These should be revisited when their dependencies are available.

---

### Recommended Next Steps

1. Split section **r** (UTF-8 B-strings) into its own RFC — it is backward-compatible and uncontroversial
2. Proceed with **a1/a2** scoped strictly to method call sites and explicit type annotations, matching C# collection expression semantics
3. Submit **e**, **h**, and **t** as independent RFCs referencing their respective suggestions

The core value of https://github.com/fsharp/fslang-suggestions/issues/1421 — reducing friction when calling .NET APIs expecting non-default numeric/collection types — is achieved through a1 and a2 without the broader inference changes.

Happypig375:
@T-Gro I understand that backward inference may be unintuitive and definitely costs compiler performance. But does this mean having `1` and `"abc"` also  work similarly to a2? i.e. forward inference is approved?
```fs
let a: byte = 1 // we know the target type is byte. Interpret 1 as a byte literal.
let b = 1 // type information for b doesn't exist yet. Infer as int.
let c: byte = b // type error.
```
```fs
let a: Span<char> = "abc" // we know the target type is char span. Interpret "abc" as char span.
let b = "abc" // type information for b doesn't exist yet. Infer as string.
let c: Span<char> = b // type error.
```

T-Gro:
Yes, for known target types this is approved.
What is not approved and was an aspect of the "q" item is the existence of `^a when ^a: [^b] and ^b: (byte|char)`, since such change would affect type checking of all string literals across all codebases.

Happypig375:
This RFC has been split into the above referenced 5 RFCs.

PR #802, opened 2025-06-10
RFC FS-1329 - Allow typed bindings(let!, use!, and!) in CE without parentheses.

edgarfgp:


RFCs/FS-1329-Allow typed bindings in CE let!s without parentheses.md
```diff
@@ -0,0 +1,283 @@
...
+Tuple patterns with type annotations require parentheses (same as regular `let`):
+
+```fsharp
+// Before (and still required after)
+async {
+    let! (a, b): int * string = asyncTuple()
+    and! (x, y): float * bool = asyncPair()
+}
+
+// After - parentheses still required (consistent with regular let)
```
T-Gro, line 98: How about tuples with each element having direct type annotation?

`let (x:int,y:int) = (5,3)` equivalent

I assume the same as regular let as well?

edgarfgp: Yes. this wont be allowed. Will update it mention this.

T-Gro (approved):

PR #805, opened 2025-07-11
[RFC FS-1151] Record spreads

brianrourkeboll:
RFC for the record-to-record spreads subset of the approved language suggestion https://github.com/fsharp/fslang-suggestions/issues/1253.

- Discussion thread (covers more than just records): https://github.com/fsharp/fslang-design/discussions/806
- Implementation: https://github.com/dotnet/fsharp/pull/18927

### DIFF
--- a/RFCs/FS-1151-record-spreads.md
+++ b/RFCs/FS-1151-record-spreads.md
@@ -824,9 +824,6 @@ This proposal combines these and adds target-directed functional _construction_ 
 | `...` occurs in an unsupported construct, such as a list, array, sequence/computation expression, or constructor-shaped record expression | **FS3902 — error:** "Spreading is not supported in this construct." |
 | A spread is put before `with`, as in `{ ...r with A = x }` or `{\| ...r with A = x \|}` | **FS3903 — error:** "Spreading is not supported in this position. Use one of the forms `{ ...expr1; A = expr2 }` or `{ expr1 with A = expr2 }` instead." |
 | A normal copy-and-update already has a `with` source and its update fields also contain a spread, e.g. `{ r with ...s }` | **FS3904 — error:** "Spread expressions and 'with' cannot be used together in the same copy-and-update expression." |
-| In a record type definition, a spread occurs to the right of another spread that contributes a field with the same name | **FS3905 — informational warning:** "Spread field '…' from type '…' shadows a field with the same name from an earlier spread." |
-| In a record type definition or a nominal or anonymous record expression, an explicit field occurs to the right of a spread that contributes a field with the same name | **FS3906 — informational warning:** "Explicit field '…' shadows a field with the same name from an earlier spread." |
-| In a nominal or anonymous record expression, a spread occurs to the right of another spread that contributes a field with the same name | **FS3907 — informational warning:** "Spread field '…' shadows a field with the same name from an earlier spread." |
 
 ## Tooling
 
--- a/RFCs/FS-1151-record-spreads.md
+++ b/RFCs/FS-1151-record-spreads.md
@@ -1,0 +1,910 @@
+# F# RFC FS-1151 - Record spreads
+
+The design suggestion [Spread operator for F#](https://github.com/fsharp/fslang-suggestions/issues/1253) has been marked "approved in principle."
+
+This RFC covers the detailed proposal for **record-to-record** spreads as the first independently useful subset of the suggestion.
+
+- [x] [Spread operator for F#](https://github.com/fsharp/fslang-suggestions/issues/1253)
+- [x] Approved in principle
+- [x] [Implementation](https://github.com/dotnet/fsharp/pull/18927)
+- [x] [Discussion](https://github.com/fsharp/fslang-design/discussions/806)
+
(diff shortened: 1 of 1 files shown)


RFCs/FS-1151-spread-operator.md
```diff
@@ -0,0 +1,14 @@
+# F# RFC FS-1151 - Spread operator
+
+The design suggestion [Spread operator for F#](https://github.com/fsharp/fslang-suggestions/issues/1253) has been marked "approved in principle."
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] [Spread operator for F#](https://github.com/fsharp/fslang-suggestions/issues/1253)
+- [x] Approved in principle
+- [ ] [Implementation](https://github.com/dotnet/fsharp/pull/TODO)
```
T-Gro: Point this at the implementation PR:

```suggestion
- [ ] [Implementation](https://github.com/dotnet/fsharp/pull/18927)
```

RFCs/FS-1151-spread-operator.md
```diff
@@ -0,0 +1,14 @@
...
+- [x] [Discussion](https://github.com/fsharp/fslang-design/discussions/806)
+
+# Summary
+
+We add support in the following specific contexts for spread expressions of the form `...source`, wherein the symbol `...` prefixed to a `source` expression or type indicates that a set of values or members is to be copied or "spread" from the value or type represented by the source expression into the target context in some way.
```
T-Gro: This needs a body before it can leave draft. Minimal spine below: Summary/scope, Motivation, Detailed design with the shadowing example, the grammar delta for the spec, and Compatibility. I tightened your Summary line in the process. Deliberately skipped Drawbacks/Alternatives (one-liners at most — `..` vs `...` is settled in #806) and the full diagnostics list (it's in #18927 if you'd rather enumerate it).

```suggestion
A spread `...source` copies the fields of a record value or type into another record. This RFC adds spreads, under `--langversion:preview`, in record type definitions — `type T = { ...U; ... }` — and in nominal and anonymous record expressions — `{ ...src; ... }` and `{| ...src; ... |}`. Spreading non-record types, mixing `...` with `with`, and spreads in collection, computation, and pattern positions are out of scope.

# Motivation

`with` copy-and-update cannot change a field's type or combine more than one source, and record types must repeat shared fields explicitly. Spreads remove that duplication.

# Detailed design

`...source` copies the fields of `source` into the surrounding record: inlined field declarations in a type definition, and field assignments in an expression, where the source is evaluated once. Fields combine left to right and a later field shadows an earlier field of the same name; a spread field shadowing an explicit field warns, and two explicit fields of the same name are an error.

    type R1 = { A: int; B: int }
    type R2 = { A: string; ...R1 }  // { A: int; B: int } — spread shadows explicit A (warns)
    type R3 = { ...R1; A: string }  // { A: string; B: int } — explicit shadows spread A

# Changes to the F# spec

A record field declaration and a record/anonymous-record expression binding each gain a spread form:

    recd-field-decl  +=  '...' type      // in record type definitions
    recd-binding     +=  '...' expr      // in record and anonymous record expressions

# Compatibility

Not a breaking change: the syntax is new and rejected by earlier compilers. Emitted assemblies contain ordinary record definitions and construction; no FSharp.Core or runtime support is required.
```

brianrourkeboll:
@dsyme This is ready for review.

Happypig375:
```fs
type A = { a : int; b : int }
type B = { a : int } with member this.b = this.a + 1
let b = { a = 1 }
let a : A = { ...b } // should this work? how about properties brought in via type extension?
let c = {| ...b |} // and then this ignores properties?
```

brianrourkeboll:
> ```fsharp
> type A = { a : int; b : int }
> type B = { a : int } with member this.b = this.a + 1
> let b = { a = 1 }
> let a : A = { ...b } // should this work? how about properties brought in via type extension?
> let c = {| ...b |} // and then this ignores properties?
> ```

https://github.com/fsharp/fslang-design/blob/9dc80f788095eb2eb9fbc0f0f97f15580c30a6ea/RFCs/FS-1151-record-spreads.md#L245

And https://github.com/fsharp/fslang-design/discussions/806#discussioncomment-17135455.

dsyme:
This is fabulous, amazing detail and close working.

One thing is this:

```
let a : Target      = { ...source }   // drops Source.Extra silently
```

I think maybe this is worth a warning, as I understand it there isn't curently a warning for this?

dsyme:
Besides that I approve this - excellent work

brianrourkeboll:
> One thing is this:
> 
> ```
> let a : Target      = { ...source }   // drops Source.Extra silently
> ```
> 
> I think maybe this is worth a warning, as I understand it there isn't curently a warning for this?

There is not currently a warning for this, no. One reason why not is that I can imagine many scenarios where that behavior would be desired — it replaces the need for explicit subtraction (`without`) when copying from a wider record to a narrower record, for example.

But we could emit an informational warning there if you think that would be helpful.

T-Gro (approved):

PR #807, opened 2025-07-23
Create FS-1152-For-loop-with-accumulation.md

Happypig375:

### DIFF
--- a/RFCs/FS-1152-For-loop-with-accumulation.md
+++ b/RFCs/FS-1152-For-loop-with-accumulation.md
@@ -1,0 +1,143 @@
+# F# RFC FS-1152 - For loop with accumulation
+
+The design suggestion ["for-with" syntactic sugar for folds](https://github.com/fsharp/fslang-suggestions/issues/1362) has not yet been marked "approved in principle".
+
+This RFC covers the detailed proposal for this suggestion.
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/1362)
+- [ ] Approved in principle
+- [ ] [Implementation](https://github.com/dotnet/fsharp/pull/FILL-ME-IN)
+- [ ] [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)
+
+# Summary
+
+A new syntax for `for`-loops that allow accumulation is introduced.
+```
+for <pat> = <expr> with <pat> in <expr> do
+    <expr>
+```
+The value of the loop body is used to update the accumulator which is returned in the end.
+
+# Motivation
+
+Currently, to do this, we need an additional mutable variable. For example,
+```fs
+let mutable sentence = "" // mutable accumulator
+for word in ["Hello", " ", "World", "!"] do
+    sentence <- sentence + word
+printfn "%s" sentence
+```
+
+However, the mutable variable leaks outside the loop body, which is undesirable.
+Moreover, the accumulator must be mutable - which is against functional immutable semantics.
+
+It is proposed that the accumulator be embeddable into the loop itself:
+```fs
+for sentence = "" with word in ["Hello", " ", "World", "!"] do
+    sentence + word
+|> printfn "%s"
+```
+The return value of the loop body updates the accumulator.
+This offers a desirable middle ground between `for` loops that must only have side-effects (returning `unit`)
+and purely functional `fold`s that must be in lambda form.
+
+Meanwhile, folds are hard to understand and the body cannot be use computation expression functionality.
+```fs
+["Hello", " ", "World", "!"]
+|> Seq.fold (fun sentence -> sentence + word (*loop body, and yet we lose the ability to do let!*)) "" // initial state is placed last??
+|> printfn "%s"
+```
+which suggests a missed opportunity to make them more familiar to people who know `for` loops.
+
+# Detailed design
+
+For this syntax
+
+```fs
+for <pat1> = <expr1> with <pat2> in <expr2> do
+    <exprs>
+    <expr3>
+```
+
+it undergoes a simple syntactical expansion to
+
+```fs
+let mutable <accum> = <expr1>
+for <pat2> in <expr2> do
+    let <pat1> = <accum>
+    <exprs>
+    <accum> <- <expr3>
+<accum>
+```
+where `<accum>` is a compiler-generated accumulation variable that cannot be used outside the loop,
+and `<exprs>` match any number of expressions before the final expression in the loop body.
+
+This means that `<accum>` is the value of this loop. It is also the accumulator which gets updated each iteration.
+
+Note that `<exprs>` are subject to usual treatment by computation expressions including application of implicit yields,
+which is different from the final expression `<expr3>`.
+
+Therefore, this may happen:
+```fs
+let result = [
+    for sentence = "" with word in ["Hello", " ", "World", "!"] do
+        sentence
+        sentence + word
+    |> printfn "%s"
+]
+// expands to
+let result = [
+    let mutable <accum> = ""
+    for word in ["Hello", " ", "World", "!"] do
+        sentence // Implicit yield here
+        <accum> <- sentence + word // But not for the final expression of the loop body
+    <accum>
+    |> printfn "%s"
+]
+```
+
+# Drawbacks
+
+It's another syntax to learn.
+
+# Alternatives
+
+Doing nothing - we miss an opportunity to bridge regular for loops with lambda folds.
+
+There is a potential extension to folding across multiple sequences with the `and` keyword in the future -
+```fs
+for s, p = 0, 0 with t in ts and u in us do
+    s + t, p + u
+```
+
+# Compatibility
+
+Please address all necessary compatibility questions:
+
+* Is this a breaking change? No
+* What happens when previous versions of the F# compiler encounter this design addition as source code? Reject syntax.
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries? Treat them the same as for loops with mutable accumulator.
+* If this is a change or extension to FSharp.Core, what happens when previous versions of the F# compiler encounter this construct? Not applicable.
+
+# Pragmatics
+
+## Diagnostics
+
+Just what you expect from a regular for loop.
+
+## Tooling
+
+Please list the reasonable expectations for tooling for this feature, including any of these:
+
+* Debugging
+  * Breakpoints/stepping
+  * Expression evaluator
+  * Data displays for locals and hover tips
+* Auto-complete
+* Tooltips
+* Navigation and Go To Definition
+* Error recovery (wrong, incomplete code)
+* Colorization
+* Brace/parenthesis matching
+
+Just what you expect from a regular for loop.


RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Alternatives
+
+Doing nothing - we miss an opportunity to bridge regular for loops with lambda folds.
+
+There is a potential extension to folding across multiple sequences with the `and` keyword in the future -
+```fs
+for s, p = 0, 0 with t in ts and u in us do
+    s + t, p + u
+```
```
brianrourkeboll, line 111: This section should mention computation expressions: https://github.com/fsharp/fslang-suggestions/issues/1362#issuecomment-2123061400.

Happypig375:
@T-Gro Review wanted

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,272 @@
...
+]
+// expands to
+let result = [
+    let mutable <accum> = ""
+    for word in ["Hello"; " "; "World"; "!"] do
+        sentence // Implicit yield here
+        <accum> <- sentence + word // But not for the final expression of the loop body
+    <accum>
+    |> printfn "%s"
+]
```
Tarmil, line 205: This opens the way for CE constructs (in this case implicit yields) appearing in sub-expressions, which feels like a bigger can of worms that should be dealt with in a separate proposal, no?

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,272 @@
...
+
+# Alternatives
+
+Doing nothing - we miss an opportunity to bridge regular for loops with lambda folds.
+
+There is a potential extension to folding across multiple sequences with the `and` keyword in the future -
+```fs
+for s, p = 0, 0 with t in ts and u in us do
+    s + t, p + u
+```
```
Tarmil, line 232: This example gives the impression that folding over N sequences would require N accumulators, but that's not the case, for example:

```fsharp
// n-dimensional Euclidian distance between A and B

let pointACoordinates = [ 1.5; 2.; -3.4; -1.2 ]
let pointBCoordinates = [ -3.1; 3.1; 1.; -0.2 ]

let euclidianDistance =
    for acc = 0. with an in pointACoordinates and bn in pointBCoordinates do
        acc + (an - bn) ** 2.
    |> sqrt
```

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,272 @@
...
+
+# Summary
+
+A new syntax for `for`-loops that allow accumulation is introduced.
+```
+for <pat> = <expr> with <pat> in <expr> do
+    <expr>
+```
+The value of the loop body is used to update the accumulator which is returned in the end.
+
```
Tarmil, line 20: `for <pat> in <expr> with <pat> in <expr> do` runs the risk of being a pretty long line, we should define nice ways to break it up into multiple lines. I propose allowing `with` to be aligned to the left with `for`:

```fsharp
for <pat> = <expr>
with <pat> in <expr> do
    <expr>
```

RFCs/FS-1152-For-loop-with-accumulation.md
Happypig375: It's no different from a regular for loop in this case. As noted in unresolved question below, there are 4 choices when dealing with implicit yields. If one finds the implicitness too dangerous, it can be changed to explicit yield required inside this kind of loop with ignore + warning without yield.

Tarmil: The problem is not the specifics of what's in the `for` loop; the problem is that the `for` loop itself is nested in a sub-expression. It's a bit obfuscated by the use of the pipe, but written differently, it looks like this:

```fsharp
let result = [
    printfn "%s" (
        for sentence = "" with word in ["Hello"; " "; "World"; "!"] do
            yield sentence
            sentence + word
    )
]
```

This is currently not possible for other CE constructs. Anything that is in a sub-expression is a normal expression, not a CE construct. For example this is not allowed:

```fsharp
async {
    printfn "%s" (
        let! x = getXAsynchronously()
        x
    )
}
```

nor this:

```fsharp
[
    printfn "%s" (
        yield 42
        "Hello world"
    )
]
```

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,277 @@
...
+Currently, to do this, we need an additional mutable variable. For example,
+```fs
+let mutable sentence = "" // mutable accumulator
+for word in ["Hello"; " "; "World"; "!"] do
+    sentence <- sentence + word
+printfn "%s" sentence
+```
+
+However, the mutable variable leaks outside the loop body, which is undesirable.
+Moreover, the accumulator must be mutable - which is against functional immutable semantics.
```
brianrourkeboll, line 32: I think recursive functions should be mentioned here (I do see a brief mention in the suggestion itself, but not here).

```fsharp
let sentence =
    let rec build sentence = function
        | [] -> sentence
        | word :: words -> build (sentence + word) words

    build "" ["Hello"; " "; "World"; "!"]

printfn $"{sentence}"
```

Recursive functions can also give more control than this proposed new construct would, e.g.:

```fsharp
let sentence =
    let rec build (sentence : StringBuilder) (words : string list) =
        match words with
        | [] -> string sentence // Do something different in the last iteration.
        | word :: words -> build (sentence.Append word) words

    build (StringBuilder ()) ["Hello"; " "; "World"; "!"]

printfn $"{sentence}"
```

RFCs/FS-1152-For-loop-with-accumulation.md
Happypig375: This should be in a different suggestion - see https://github.com/dotnet/fsharp/issues/18754#issuecomment-3101793041

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,474 @@
...
+
+However, the mutable variable leaks outside the loop body, which is undesirable.
+
+Moreover, the accumulator must be mutable - which is against functional immutable semantics.
+The danger with mutable stateful objects is that it further encourages globally mutable state against functionally immutable design. This is why `fold` exists: to encapsulate mutability and temporary variables as well.
+
+It is proposed that the accumulator be embeddable into the loop itself, with a familiar and sensible parameter order the same as the fold lambda body, i.e. accumulator first, sequence second, such that refactors from `fold`s are easy:
+```fs
+for sentence = "" with word in ["Hello"; " "; "World"; "!"] do
+    sentence + word
```
charlesroddie, line 39: This syntax makes no sense because `sentence + word` is not something that can be "done": it's a string and a string cannot be "done". `for` makes no sense in this context since `sentence=""` indicates an intention for sentence to take only one value, namely "", so there is nothing that `for` iterates over.

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,474 @@
...
+```
+The return value of the loop body updates the accumulator. This is a synthesis of `fold`s with loops, therefore it can be called a "fold loop". There is no direct equivalent in other languages, they either have `fold` with a lambda or `for` loops with no accumulator.
+
+This offers a desirable middle ground between `for` loops that must only have side-effects (returning `unit`)
+and purely functional `fold`s that must be in lambda form. It is declarative which removes concerns about order of operations. It encourages functional pureness without the shortcomings of `fold`.
+
+`fold` is inferior to this syntax because it is hard to understand.
+```fs
+["Hello"; " "; "World"; "!"]
+|> Seq.fold (fun sentence -> sentence + word (*loop body but shown as a lambda!*)) "" // initial state is placed last??
```
charlesroddie, line 50: fold does not require placing the data last so this comment is inaccurate

RFCs/FS-1152-For-loop-with-accumulation.md
```diff
@@ -0,0 +1,474 @@
...
+    sentence <- sentence + word
+printfn "%s" sentence
+```
+
+However, the mutable variable leaks outside the loop body, which is undesirable.
+
+Moreover, the accumulator must be mutable - which is against functional immutable semantics.
+The danger with mutable stateful objects is that it further encourages globally mutable state against functionally immutable design. This is why `fold` exists: to encapsulate mutability and temporary variables as well.
+
+It is proposed that the accumulator be embeddable into the loop itself, with a familiar and sensible parameter order the same as the fold lambda body, i.e. accumulator first, sequence second, such that refactors from `fold`s are easy:
```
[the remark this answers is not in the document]
reinux: I do wonder if maybe it should have the `with` come after and mark the accumulator instead, i.e.
```fsharp for i in 1..10 with x do ...```

File: RFCs/FS-1152-For-loop-with-accumulation.md (hunk 7 above)
Comment (Happypig375) [reply RFCs/FS-1152-For-loop-with-accumulation.md]: Updated it

File: RFCs/FS-1152-For-loop-with-accumulation.md (hunk 6 above)
Comment (Happypig375) [reply RFCs/FS-1152-For-loop-with-accumulation.md:39]: > a string cannot be done

```fs
[ for word in ["Hello"; " "; "World"; "!"] do word ]
```
A string can still be done, just yielded implicitly. The same logic applies here - "do" this evaluation for the loop iteration.

> `sentence=""` indicates an intention for sentence to take only one value

```fs
for i = 1 to 10 do printfn $"{i}"
```
initializing loop state with `=` is already existing syntax for the integer for loop.

Happypig375:
@charlesroddie It's an RFC for a non-approved suggestion. The README explicitly states that doing this and starting a prototype implementation can help convince.

Smaug123:
(By the way, you could link to https://github.com/Happypig375/fslang-design/blob/patch-29/RFCs/FS-1152-For-loop-with-accumulation.md directly in the PR description if you want to save people a couple of clicks.)

RFCs/FS-1152-For-loop-with-accumulation.md
Happypig375: Wrote this clarification in the RFC.

RFCs/FS-1152-For-loop-with-accumulation.md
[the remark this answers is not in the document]
T-Gro: I would not consider order of existing function parameters relevant at all - the choice of function parameters in fslib is done with different set of criteria (partial application, piping)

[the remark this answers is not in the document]
Happypig375: I adopted the parameter order of iteration item first and accumulator later. Seems everyone prefers it - and I have included the semantic deduction process in the RFC.

thinkbeforecoding:
The usage would be something like this (to avoid mutation, and make for an expression) ?
```fsharp
let total =
    for item in items with x = 0 do
        x + item.Count
```

RFCs/FS-1152-For-loop-with-accumulation.md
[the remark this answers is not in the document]
Happypig375: The resolution on initialize omission is to not support it. Justification at 78d28e34f25557a56d5f78020d34ee638099551c

Happypig375:
@T-Gro How do you feel about the alternatives to choose (value returning vs binding leaking) in the proposal?

T-Gro:
> @T-Gro How do you feel about the alternatives to choose (value returning vs binding leaking) in the proposal?

This should come with the standard `result ignored` warning, like it applies to all expressions.
This will be tricky to explain clearly - even though it syntactically builds upon `for`, it really changes it from a statement (unit-returning) to an expression.

T-Gro:
They should be treated both as valid, I don't think there is a big difference.

Happypig375:
But treating `->` the same as `do` for value-returning version would conflict with the existing meaning of `do yield` in CEs.

Happypig375:
@omcnoe If you really think about it, the traditional `for` loop just has implicit `with state = ()`, i.e. accumulates to unit. If you change the accumulation, you also change the loop value. They aren't actually independent.

Happypig375:
@omcnoe See if https://github.com/fsharp/fslang-design/pull/807/commits/23a218ce048321294c5550cad1af12c162585b6b explains it well

Martin521:
> Also, I don't get why you are complaining about procedures instead of the actual suggestion - it seems not to be constructive behaviour especially when paired with the 👎 reaction on the top post, as seen not just in this PR, but also in other issues, where the 👎 reaction persists despite a proper reply.

I do think that the language design process makes sense and I therefore believe it is ok if somebody points to it.
Also, there can be many reason to not favor the addition of a feature even after "proper replies".
My own opinion in this case: The feature is somehow cool, but I still don't want it in F#, based on my perception of the cost-benefit ratio. And because I believe that for the future of F# it is much more important to focus on a) open issues that bother the average user and b) on maintainability of the compiler.

krauthaufen:
In my opinion this construct would make it more complex for newcomers to understand code, intransparent about what is actually executed underneath (Seq.fold, List.fold, a for loop with a mutable variable?) and I don't see how it makes code shorter or easier to read. A for loop with a mutable variable is a few characters longer and familiar to most people. Folds are also more transparent in my opinion and can be used in pipe-chains.
I'm just concerned about F#'s simplicity being eaten up by these corner-case specializations.

reinux:
@Happypig375 did establish that [folds aren't an edge case by any stretch](https://github.com/fsharp/fslang-design/commit/0135550679daf3fca213eed877338ead2a95c6c3) -- in fact, it's the third most common HOF after `map` and `iter`, both of which can be expressed as list/seq comprehensions and bare for loops respectively, and also can't be used in a pipe. `fold` is used 90% as often as `iter` and more than twice as often as `choose`, the next most common HOF. Certainly, folds are used _a lot_ in elm-style programs, as you often need to update the state over a sequence.

To me, the main downside of using a mutable is that, especially as beginners, we're being encouraged to use `mutable` as sparingly as possible moreso for the sake of discipline than to avoid any immediate danger of juggling state in a brief loop. And then turning around and instructing them to use mutables as an alternative for something as common as folds sends a mixed message -- something I still remember being a head-scratcher 15+ years ago when I first started with functional programming in F#. It _greatly_ simplifies tutorials and documentation if we don't need to make this concession.

Likewise, the downside of folds, especially for beginners, is that they require memorizing a lot of seemingly arbitrarily ordered arguments: `('S -> 'T -> 'S) -> 'S -> 'T seq -> 'S` is pretty overwhelming.

Virtually the only situation you'd ever reach for the `||>` operator is to alleviate the problem with folds obscuring the `state` and `xs` args behind a giant function argument with potentially several nested brackets. Although it's a dead simple one-liner in the core library as opposed to a language feature as in this proposal, as far as the user is concerned, a new operator is _always_ new syntax. And it flips the argument order and switches from curried to tupled.

Finally, this is a relatively minor point, but if you're trying to draft up a loop or HOF application, you might first try a fold and find that it isn't sufficient. I think both beginners and advanced users alike have this experience routinely. In those situations, refactoring the function argument to an HOF is not as intuitive as adding or removing things to/from a loop.

krauthaufen:
Just to clarify, I didn‘t mean to say that folds are not common, I meant to say that the places where you actually benefit from this syntax (meaning shorter or more readable code) seem pretty far fetched to me. The „don‘t use mutables“ argument is actually pretty bad since mutable variables inside a function body don’t harm anyone. Global mutables and mutable objects kill functional programming, not some accumulator variable.

reinux:
I wouldn't argue that it's shorter. F# doesn't really go out of its way to make code shorter. It may or may not be more readable to an experienced user either.

The main point is that it can help with adoption if users don't have to go through the whole journey of "Wow, folds are pretty overwhelming, can't I just use a loop? Oh I guess I can but can I do it without using a mutable since I'm trying to practice immutability right now? Oh okay I guess I can use `||>` but it's not much of an improvement... I guess I'll stick with folds"

krauthaufen:
I don’t see how this should be beginner friendly, you use an imperative construct (a loop) which in its body returns a value which is initialized by using `with s = 0` (a keyword suggesting an immutable declaration) where in the loop-body it‘s actually replacing its value with the new one provided and in the end your entire loop returns the overall value. I don‘t claim to have a Solution for your syntactic issues with fold but maybe a simple CE builder for this would also do the job?

reinux:
Although it's in a different form, intuitively, it feels a lot like the `for` loops in C-style languages, with the initialization, the condition and the increment. It's been a while, but I don't remember C for loops ever confusing me. I think it's a pretty big improvement in beginner friendliness. I find that formalized imperative styles are often a good compromise between safety and ease of learning, which is often why list comprehensions are preferred over maps.

The main problem with CEs, which has been pointed out in the RFC, is that they're notoriously difficult to debug.

T-Gro:
Thank you for putting together this detailed RFC. The level of detail provided was genuinely helpful in understanding both the motivation behind the proposal and the various technical options considered. It's clear that significant thought and effort went into this document.

After careful consideration, I've decided to close this PR for now. The reasoning is captured in the discussion thread, where I've marked the suggestion as ["Probably not"](https://github.com/fsharp/fslang-suggestions/issues/1362#issuecomment-3580698414).

As noted there, this doesn't preclude revisiting the idea in the future should circumstances change—whether through new compelling use cases, shifts in the ecosystem, or renewed community interest and consensus.

Thank you again for your contribution to the F# language design process.

PR #809, opened 2025-08-03
create RFC  FS-1330 - Support tailcalls in computation expressions

majocha:


RFCs/FS-1330-support-tailcalls-in-computation-expressions.md
```diff
@@ -0,0 +1,117 @@
...
+Please address all necessary compatibility questions:
+
+* Is this a breaking change?
+> Possibly, if a builder already defines `ReturnFromFinal`/`YieldFromFinal` methods.
+
+* What happens when previous versions of the F# compiler encounter this design addition as source code?
+>N/A
+
+* What happens when previous versions of the F# compiler encounter this design addition in compiled binaries?
+>N/A
```
T-Gro, line 85: CE authors should be aware that older compilers (older SDKs) will not call the `FromFinal` methods.
The CE should work in a `progressive enhancement` kind of way - making sure it stays correct even when the `FromFinal` is not called.

T-Gro (approved):

PR #814, opened 2025-08-12
Create FS-1332 - `assert` keyword enhancement

ijklam:

### DIFF
--- a/RFCs/FS-1332-assert-keyword-enhancement.md
+++ b/RFCs/FS-1332-assert-keyword-enhancement.md
@@ -1,6 +1,6 @@
 # F# RFC FS-1332 - `assert` keyword enhancement
 
-The design suggestion [`assert` keyword: have compiler emit call to `Debug.Assert` overload with CallerArgumentExpression where possible](https://github.com/dotnet/fsharp/issues/18489) has been marked "approved in principle".
+The design suggestion [`assert` keyword: have compiler emit call to `Debug.Assert` overload with CallerArgumentExpression when possible](https://github.com/dotnet/fsharp/issues/18489) has been marked "approved in principle".
 
 This RFC covers the detailed proposal for this suggestion.
 
@@ -15,15 +15,14 @@ When an `assert` expression failed, it will shows the text of the expression in 
 
 # Motivation
 
-When the `assert` expression failed, it does not provide the details of the expression, but the plain `System.Diagnostics.Debug.Assert` does.
+When the `assert` expression failed, it does not provide the details of the expression, but the `System.Diagnostics.Debug.Assert` does.
 
 # Detailed design
 
 `assert <bool expression>` will be translated into `System.Diagnostics.Debug.Assert(<bool expression>, "<bool expression>")`.
 
 ```fsharp
-// The below 2 lines are equivalent
-assert (1 + 1 = 2)
+assert (1 + 1 = 2) // This will be translated into
 System.Diagnostics.Debug.Assert((1 + 1 = 2), "(1 + 1 = 2)")
 ```
 
@@ -70,7 +69,7 @@ Please address all necessary compatibility questions:
 
 ## Performance
 
-This feature may impact the performance when the code file is too large and has many `assert` expressions, since it needs get substrings from the file.
+This feature may impact the compilation speed when the code file is too large and has many `assert` expressions, since it needs get substrings from the file.
 
 ## Scaling
 


RFCs/FS-1332-assert-keyword-enhancement.md
```diff
@@ -0,0 +1,84 @@
+# F# RFC FS-1332 - `assert` keyword enhancement
+
+The design suggestion [`assert` keyword: have compiler emit call to `Debug.Assert` overload with CallerArgumentExpression where possible](https://github.com/dotnet/fsharp/issues/18489) has been marked "approved in principle".
```
T-Gro, line 3: What does `where possible` mean here?

RFCs/FS-1332-assert-keyword-enhancement.md
```diff
@@ -0,0 +1,84 @@
...
+- [x] [Implementation](https://github.com/dotnet/fsharp/pull/17519)
+- [ ] [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)
+
+# Summary
+
+When an `assert` expression failed, it will shows the text of the expression in the exception message.
+
+# Motivation
+
+When the `assert` expression failed, it does not provide the details of the expression, but the plain `System.Diagnostics.Debug.Assert` does.
```
T-Gro, line 18: What does `plain ... does` mean here?

RFCs/FS-1332-assert-keyword-enhancement.md
```diff
@@ -0,0 +1,84 @@
...
+* What happens when this feature is consumed by another .NET language?
+    > It will works the same as before.
+* Are there any planned or proposed features for another .NET language (e.g., [C#](https://github.com/dotnet/csharplang)) that we would want this feature to interoperate with?
+    > N/A
+
+# Pragmatics
+
+## Performance
+
+This feature may impact the performance when the code file is too large and has many `assert` expressions, since it needs get substrings from the file.
```
T-Gro, line 73:
```suggestion
This feature may impact the compilation speed when the code file is too large and has many `assert` expressions, since it needs get substrings from the file.
```

With performance, please always differentiate between compile time / design time / runtime ...

RFCs/FS-1332-assert-keyword-enhancement.md
ijklam: That is the original title

RFCs/FS-1332-assert-keyword-enhancement.md
```diff
@@ -0,0 +1,84 @@
...
+When the `assert` expression failed, it does not provide the details of the expression, but the plain `System.Diagnostics.Debug.Assert` does.
+
+# Detailed design
+
+`assert <bool expression>` will be translated into `System.Diagnostics.Debug.Assert(<bool expression>, "<bool expression>")`.
+
+```fsharp
+// The below 2 lines are equivalent
+assert (1 + 1 = 2)
+System.Diagnostics.Debug.Assert((1 + 1 = 2), "(1 + 1 = 2)")
```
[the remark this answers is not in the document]
ijklam: Yes but this is to say the first line will be translated into the second line. Not all things are equivalent.

RFCs/FS-1332-assert-keyword-enhancement.md
ijklam: The expression updated

Martin521 (changes requested):
I am not sure if a RFC makes sense here. The mentioned suggestion is not a design suggestion, but an issue in the compiler repo. The proposal there is to change the implementation of the [assert expression](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/assertions) (with the help of `CallerArgumentExpression`) such that the failure message contains the assert condition, similar to what C# offers. 

That would be an improvement of the assert message that does not need a spec change. Once there is agreement in the issue thread that it is useful, and `CallerArgumentExpression` becomes available, it can simply be implemented by a PR in the compiler repo.

If the intent behind this RFC is something different, then I guess it should start with a design suggestion in the fslang-design repo.

brianrourkeboll:
> I am not sure if a RFC makes sense here. The suggestion it refers to is not a design suggestion, but an issue in the compiler repo. The proposal there is to change the implementation of the [assert expression](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/assertions) (with the help of `CallerArgumentExpression`) such that the failure message contains the assert condition, similar to what C# offers.

I guess this does technically change the spec, though:

https://github.com/fsharp/fslang-design/blob/d2956ef51d9de03ddcca625a4c1265a360d897ec/RFCs/FS-1332-assert-keyword-enhancement.md?plain=1#L31-L33

Martin521:
> I guess this does technically change the spec, though:

True. I missed that.
Not sure if this warrants a RFC. If so, it can be very short, just the "Summary" and "Changes to the F# spec" sections.

T-Gro (approved):

T-Gro:
> I am not sure if a RFC makes sense here. 

It's living on a boundary - it is a compiler intrinsic, acting as a defacto part of the language.
The RFC allows us to agree on the feature itself incl. its details 👍

PR #830, opened 2026-02-22
Add FS-1337 - Ignore file-based-app directives

Martin521:

Ignore lines starting with `#:` to allow experimentation with F# file-based-apps. ([Suggestion #1440](https://github.com/fsharp/fslang-suggestions/issues/1440))


T-Gro (approved):
I gave this a new thought and it makes a lot more sense to me to enable experimentation without longer waits.

People could then keep around a stable version of the compiler, and just experiment on the tooling layer.


One particular experiment I recently thought of was "file-based apps with multi-language `#load` (or whatever syntax) support.". I.e. a mixed language scenario without explicit web of projects in a solution.

Given some prototyping and playing around, this could also resolve some longer standing needs around more elegantly supporting source generators that consume C# AST and produce C# as a string. (F# and C# still in different files, but at least without the project overhead)

(yes, this might have been a nerd sniping comment)

PR #309, opened 2018-06-01
Update FST-1005 for new requirements for Microsoft employees

cartermp:
There is a new policy for Microsoft employees w.r.t publishing assets to NuGet. This is really just about cleaning things up on our end. This does not affect the F# Software Foundation's ownership of various packages.

PR #832, opened 2026-04-27
RFC FS-1337: Optional Automatic File Order

bryancostanich:
Drafting RFC FS-1337 covering [dotnet/fsharp#19647](https://github.com/dotnet/fsharp/pull/19647) (opt-in `--file-order-auto+`).

This is filed as a **draft** because the auto-inference path doesn't yet have its own `approved-in-principle` target. [Suggestion #309](https://github.com/fsharp/fslang-suggestions/issues/309) covers an adjacent direction (explicit TypeScript-style imports). I've left a comment on the implementation PR (dotnet/fsharp#19647) asking whether #309's scope should be split into two approval tracks (auto-inference + explicit imports), or whether a brand-new fslang-suggestion should be filed for the auto-inference path.

The two are complementary, not competing. Posting the RFC now so reviewers have something concrete to react to when deciding which approval target this should anchor on.

## Summary

`--file-order-auto+` is an opt-in compiler flag (off by default) that lets the compiler topologically sort source files from their declarations and qualified-identifier references. No new syntax. Cycles get wrapped in a synthesised `namespace rec` on the build path. Validated against 7 real OSS F# projects (Argu, FsCheck, FSharpPlus, FsToolkit.ErrorHandling, Expecto, FSharp.Data.Json.Core, Fable.Promise) — auto-mode adds zero errors over baseline on every buildable target.

See the RFC document for full design, alternatives, drawbacks, and validation evidence.

### DIFF
--- a/RFCs/FS-1337-optional-automatic-file-order.md
+++ b/RFCs/FS-1337-optional-automatic-file-order.md
@@ -1,0 +1,194 @@
+# F# RFC FS-1337 - Optional Automatic File Order (`--file-order-auto+`)
+
+- [x] [Suggestion: Syntactically describe dependencies between files (#309)](https://github.com/fsharp/fslang-suggestions/issues/309), see Alternatives
+- [ ] Approved in principle
+- [x] [Implementation: PR #19647](https://github.com/dotnet/fsharp/pull/19647)
+- [ ] Discussion
+
+## Summary
+
+Add an opt-in compiler flag, `--file-order-auto+` (off by default), that lets the compiler figure out the file order itself by reading what each file declares and what it references. List your source files (impl `.fs` and signature `.fsi`) in any order in the `.fsproj`, and the compiler topologically sorts them before type checking. Files that mutually reference each other get wrapped in a synthesised `namespace rec` on the build path, so cross-file mutual recursion just works without `and`-chains. Wire it through MSBuild with `<FSharpAutoFileOrder>true</FSharpAutoFileOrder>`. F# Compiler Service (FCS) hosts (Ionide and friends) opt in via `FSharpProjectOptions.OtherOptions`. F# Interactive (FSI) is not targeted by this RFC; multi-file `#load` semantics are untouched.
+
+The flag changes nothing about type inference, module/namespace semantics, accessibility, signature files, or `module rec`/`namespace rec`. With the flag off, this is byte-identical to upstream.
+
+## Motivation
+
+I love F#. I've also bounced off using it more, for years, mainly because of one thing: having to maintain a topological order of source files in `.fsproj`, and the resulting reliance on `and`-joined recursive type chains to work around it. It's accidental complexity in a language that otherwise does so much right.
+
+Issue [#309](https://github.com/fsharp/fslang-suggestions/issues/309) has been tracking this whole space since 2014, with a bunch of distinct proposals: explicit file-to-file imports (TypeScript-style), `#require` directives, `fileorder.txt`, project-file changes. They all share a goal (get rid of the manual ordering tax) and they all require some user-visible change (new syntax, new files, manual migration).
+
+This RFC takes the lowest-friction path: opt in, and the compiler figures it out. No new syntax, no edits to source files, no new project conventions beyond a single boolean property. People who want the pain to go away flip the property; people who want explicit control keep manual ordering, or layer a future explicit-import feature on top.
+
+The framing matches what @dsyme and @nojaf were sketching in #309 [in 2022](https://github.com/fsharp/fslang-suggestions/issues/309#issuecomment-1290876076): an opt-in mechanism that, when adopted in large codebases, unlocks parallelization and finer-grained incremental checking. This RFC delivers that mechanism via auto-inference. An explicit-import RFC could deliver the same mechanism via manual annotation. Both are useful for different audiences, and they shouldn't compete.
+
+## Design
+
+The flag inserts a dependency-ordering pass between parse and check:
+
+```
+parsedInputs ──▶ [ enter phase: stub TcEnv ] ──▶
+              ──▶ [ symbol collection: extract decls + refs ] ──▶
+              ──▶ [ dep graph + Tarjan SCC ] ──▶
+              ──▶ [ apply file order, synthesise cycle groups ] ──▶
+              ──▶ check
+```
+
+Three sub-systems do the work:
+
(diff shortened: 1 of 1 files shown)


RFCs/FS-1337-optional-automatic-file-order.md
```diff
@@ -0,0 +1,194 @@
+# F# RFC FS-1337 - Optional Automatic File Order (`--file-order-auto+`)
+
+- [x] [Suggestion: Syntactically describe dependencies between files (#309)](https://github.com/fsharp/fslang-suggestions/issues/309), see Alternatives
+- [ ] Approved in principle
+- [x] [Implementation: PR #19647](https://github.com/dotnet/fsharp/pull/19647)
+- [ ] Discussion
+
+## Summary
+
+Add an opt-in compiler flag, `--file-order-auto+` (off by default), that lets the compiler figure out the file order itself by reading what each file declares and what it references. List your source files (impl `.fs` and signature `.fsi`) in any order in the `.fsproj`, and the compiler topologically sorts them before type checking. Files that mutually reference each other get wrapped in a synthesised `namespace rec` on the build path, so cross-file mutual recursion just works without `and`-chains. Wire it through MSBuild with `<FSharpAutoFileOrder>true</FSharpAutoFileOrder>`. F# Compiler Service (FCS) hosts (Ionide and friends) opt in via `FSharpProjectOptions.OtherOptions`. F# Interactive (FSI) is not targeted by this RFC; multi-file `#load` semantics are untouched.
```
dsyme, line 10: Placing files in a *rec" group may mess with the initialization semantics for let (and static let) bindings in the files. This needs real care to make sure it isn't possible to ever see uninitialized let values in F#

RFCs/FS-1337-optional-automatic-file-order.md
```diff
@@ -0,0 +1,194 @@
...
+
+`.fsi`/`.fs` pairs are collapsed to one logical contributor for export-map purposes. A sig→impl edge redirect ensures consumers depending on a sig file get sorted after the impl, keeping the pair adjacent and consumers correct.
+
+### Cycle group synthesis (build path only)
+
+Files in an SCC of size > 1 are wrapped into one `ParsedImplFileInput` whose top-level `SynModuleOrNamespace` entries get `isRecursive = true`, effectively a `namespace rec` covering the original modules. This is what makes cross-file `type Tree ↔ Forest` compile without `and`-chains. `open` declarations inside the synthesised block are hoisted to the front of each module/namespace (the FS3200 fix). Cross-namespace cycle groups fall back to original order to avoid FS0247 (a synthesised `module Y` inside `namespace rec X` would conflict with the original `namespace X.Y`).
+
+FCS does not synthesise cycle groups in this RFC. IDE diagnostics for cycle-heavy projects show the cycle as a normal type error; the build path resolves it. This is called out in the migration docs. Adding FCS-level cycle synthesis is a follow-on RFC because it requires designing incremental graph invalidation.
```
dsyme, line 52: Realistically FCS must be part of this RFC - we can't have language features that don't have good tooling.

bryancostanich:
@dsyme fantastic points, i'll chew and get back to it!

auduchinok:
I'd like to second that FCS scenarios should be considered here. Moreover, thinking about how users use the IDE may change the implementation drastically.

When building a project using the compiler, the file order is not expected to change, so it sounds like calculating it based on the graph is sufficient. However, when working on a project, it's actually expected that the ordering will be changing, so I'd think about editing a project as something more important than the actual building.

Consider a case when files have to be reordered due to a refactoring made by a user. How would FCS handle that? Should a user trigger reordering in their IDE explicitly somehow? It means the tools will need to implement it too. Alternatively, FCS could do it automatically, but it sounds too unpredictable and may also worsen the performance of the IDE (due to how caches would be thrown away on the file order change). This has to be thought-through.

What if there's no correct order during a massive refactoring? How can we handle that? What if there's circular dependency? What if cutting a type during a refactoring changes the order in a way that other things resolve differently?

RFCs/FS-1337-optional-automatic-file-order.md
```diff
@@ -0,0 +1,194 @@
...
+
+### `and` keyword deprecation (FS3887)
+
+When `--file-order-auto+` is set, `and`-joined type chains emit warning FS3887 ("The 'and' keyword for mutually recursive types is unnecessary when using `--file-order-auto`. Consider placing types in separate declarations. This keyword may be removed in a future version."). Suppressable via `--nowarn:3887` or `<NoWarn>3887</NoWarn>`. Silent in manual mode. The `and` keyword itself is not deprecated globally; only its use as a workaround for cross-file ordering becomes redundant.
```
auduchinok, line 56: I think this should not be part of this design suggestion. `and` is a often used keyword in real code due to how the language works, and changes to how `fsproj` files are handled should not be related to it.

RFCs/FS-1337-optional-automatic-file-order.md
```diff
@@ -0,0 +1,194 @@
...
+
+`.fsi`/`.fs` pairs are collapsed to one logical contributor for export-map purposes. A sig→impl edge redirect ensures consumers depending on a sig file get sorted after the impl, keeping the pair adjacent and consumers correct.
+
+### Cycle group synthesis (build path only)
+
+Files in an SCC of size > 1 are wrapped into one `ParsedImplFileInput` whose top-level `SynModuleOrNamespace` entries get `isRecursive = true`, effectively a `namespace rec` covering the original modules. This is what makes cross-file `type Tree ↔ Forest` compile without `and`-chains. `open` declarations inside the synthesised block are hoisted to the front of each module/namespace (the FS3200 fix). Cross-namespace cycle groups fall back to original order to avoid FS0247 (a synthesised `module Y` inside `namespace rec X` would conflict with the original `namespace X.Y`).
```
T-Gro, line 50: Adding automatic isRecursive=true drops a safety feature of the language (preventing ACCIDENTAL recursions, leading to either inifinite loops or stackoverflows).
I think this idea is already too opinionated on its own, I would not recommend adding another dimension (i.e. dropping something seen as a feature by users) to it.

RFCs/FS-1337-optional-automatic-file-order.md
T-Gro: `and` is really a safety feature to explicitly control where recursion can or a cannot happen, just like `rec` is for the individual case.
Even though there is merit in making `fsproj` management easier, I believe it should not lead to a reduction of safety.

T-Gro:
@bryancostanich :

I think the next gap to close when it comes to feature design is to figure out incrementally - that is the tooling and F.C.S aspects @auduchinok  and @dsyme  mentioned.

The RFC should cover for it  (e.g. have a look at TransparentCompiler and IncrementalBuild in the dotnet/fsharp repo.
Can the algorithm for order derivation be modified to work incrementally on diffs ? 
If 1 file out of 100 gets changed, what is the worst case scenario for re-checking the full project?

PR #838, opened 2026-08-27
RFC FS-1342: Contextual target typing for numeric and collection literals

Happypig375:

## Summary

Adds a focused RFC for contextual target typing of unsuffixed numeric literals and fixed-element bracket collection expressions.

This is the level 1–2 subset requested in the review of #800:

- arguments to CLI methods, constructors, indexers, property setters, and known delegate invocations;
- expressions checked under an explicit target annotation;
- no extension to ordinary curried F# function application in the first version;
- no backwards inference through bindings or later uses;
- no generalized literal constraints, pipeline propagation, or generic-type inference from a literal.

Uncontextualized literals keep their existing default types. The proposal is therefore a local, top-down conversion rather than a redefinition of literal inference.

The RFC specifies:

- compatibility-first overload resolution;
- target-specific numeric parsing and compile-time range checking;
- array, `Span<_>`, `ReadOnlySpan<_>`, and `CollectionBuilderAttribute` collection targets;
- evaluation and lowering requirements;
- diagnostics and tooling;
- source and binary compatibility;
- a staged implementation and test plan.

## Related work

- Suggestion: https://github.com/fsharp/fslang-suggestions/issues/1421
- Original omnibus RFC: #800

This PR supersedes only the approved level-1 and level-2 target-typing portion of #800. It does not carry forward the rejected backwards-inference or new-SRTP-constraint designs.

`RFCs/FS-1342-target-typed-literals.md`

### DIFF
--- a/RFCs/FS-1342-target-typed-literals.md
+++ b/RFCs/FS-1342-target-typed-literals.md
@@ -24,8 +24,6 @@ let kind: byte = 3
 let data: byte array = [0; 64; 128; 255]
 ```
 
-Elsewhere, current defaults and inference remain unchanged. The feature creates no literal constraint and does not infer backwards through bindings, pipelines, returned values, or operator operands.
-
 # Motivation
 
 At .NET API boundaries, F# callers must often restate an already known parameter type with numeric suffixes, array syntax, span conversion, or collection factories. Contextual construction removes that ceremony and can avoid an intermediate F# list while retaining compile-time numeric range checks.
@@ -36,7 +34,7 @@ The restricted contexts preserve local reasoning and existing inferred signature
 
 ## Context and ordering
 
-A target is **fully known** when no unresolved inference variable relevant to the conversion remains. The literal cannot itself determine a generic argument.
+A target is **fully known** when no unresolved inference variable relevant to the conversion remains. The literal cannot itself determine a generic argument. Outside the two entry points in the summary, current defaults and inference apply; no literal constraint or backwards inference is introduced.
 
 For a non-overloaded target, first check the expression using current F# rules, including [FS-1093](../FSharp-6.0/FS-1093-additional-conversions.md). Attempt contextual literal conversion only if that fails.
 
@@ -136,12 +134,11 @@ Diagnostics list the contextual target for each ambiguous candidate.
 
 # Changes to the F# spec
 
-- **Expression checking:** add contextual-literal mode at the two approved entry points and through the listed result-propagating forms; define it as a last-resort conversion.
+- **Expression checking and method application:** add the contextual mode, propagation boundary, and two-phase ordering in [Context and ordering](#context-and-ordering).
 - **Numeric literals:** add direct target parsing for the table above without changing lexical grammar.
-- **Collection expressions:** define the eligible bracket subset, targets, element checking, construction, ref safety, and quotation behavior.
-- **Method application resolution:** add Phase A and Phase B before contextual better-conversion comparisons.
+- **Collection expressions:** adopt the syntax, target, construction, ref-safety, and quotation contracts above.
 - **Inference:** add no constraint form, generalization rule, or inferred signature.
-- **FSharp.Compiler.Service:** record whether conversion was contextual and expose selected target, element type, overload, and construction members.
+- **FSharp.Compiler.Service:** expose the selected target, element type, overload, and construction members.
 
 # Drawbacks
 
@@ -162,7 +159,7 @@ C# has constant-expression numeric conversions, target-typed collection expressi
 
 # Compatibility
 
-The Phase-A rule preserves the interpretation and overload winner of existing valid calls; ordinary literals retain current defaults. The feature makes previously invalid annotated or call-site forms valid. It emits existing CLI constants, arrays, spans, constructors, and calls, so it adds no runtime representation or FSharp.Core dependency. Older compilers reject source relying on the conversion but can consume produced assemblies, subject to their target frameworks.
+The fallback ordering in [Context and ordering](#context-and-ordering) makes previously invalid annotated or call-site forms valid without changing an existing winner. Produced assemblies use existing CLI constants, arrays, spans, constructors, and calls, so older compilers can consume them, subject to their target frameworks, and no FSharp.Core dependency is added.
 
 A later overload can change or make ambiguous a call already relying on Phase B, as with other overload versioning. The feature should initially be gated by the corresponding preview language version.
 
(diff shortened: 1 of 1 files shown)


RFCs/FS-1342-target-typed-literals.md
```diff
@@ -0,0 +1,682 @@
...
+
+# Alternatives
+
+## Keep suffixes and factories only
+
+This is maximally explicit but retains avoidable ceremony at .NET API boundaries and can force an intermediate representation.
+
+## Use only `op_Implicit`
+
+`op_Implicit` from the default source type cannot provide direct target-specific parsing for every narrowing literal, and string/list-based conversions may allocate before conversion. A general multi-step user-conversion search also introduces ambiguity and side-effect questions outside this RFC.
```
charlesroddie: Good to have this RFC. I'm putting some initial thoughts but will work them out in more detail.

It's really in spirit an extension of the `op_Implicit` RFC which was discussed and written up by Don very carefully: https://github.com/fsharp/fslang-design/blob/main/FSharp-6.0/FS-1093-additional-conversions.md

That RFC gives a series of warnings, and there is a default for what is allowed without warning, but you can also make it stricter or more relaxed by adjusting project warning settings.

This RFC is an extension of that one. One way it's an extension is that numeric conversions could happen at compile time and bypass `op_Implicit`. So via "numeric interpretation" (this RFC), in addition to the existing "op_Implicit", "upcasting" something of type `X` can be interpreted as a `Y`.

So some questions are:
- Does this work within the existing system of warnings? My initial guess is that this hasn't yet been considered but that it can do, where the numeric conversions allowed are just a more extended set than that allowed in FS-1073.
- Does it do more than just extending the set of numeric conversions and applying them at compile time? If it does, then those extensions are likely to be the most novel extensions and will have further questions about whether users will be confused that a numeric interpretation will work here but an upcast or op_Implicit can't work. If they aren't then this RFC could be quite compacted as the implications for the typing of F# expressions would be very narrowly scoped.

RFCs/FS-1342-target-typed-literals.md
```diff
@@ -0,0 +1,682 @@
...
+
+The following retain their current F# list-expression meaning in the first version:
+
+- range syntax such as `[1 .. 10]`;
+- `for`, `while`, `let`, `use`, `try`, `if`, or `match` statements at collection-expression level;
+- other sequence-expression constructs whose lowering is not a simple sequence of element and spread operations.
+
+They may still appear inside a parenthesized element expression. Extending contextual construction to the full sequence-expression grammar can be considered separately.
+
+### Target types
```
charlesroddie: This was a different language suggestion, not approved in principle. It's popular and I can see it being approved in principle in the near future but the issues involved but it's very large in scale. The rest of this PR, the numeric stuff, is small to medium but already needs a lot of work. So this definitely needs an approval in principle and a separate RFC.

If it goes to RFC issues would include:
- It's reasonable to think that this can be done to match C# behaviour. Will it match F#'s list expressions, e.g. with `for` loops, `yield`s, `yield!`s etc? I think it would be very unexpected if the answer is no.
- The other side of the this is explicit construction. We currently have state machine builder infrastructure. Could this be moved to a syntax like `list [ ... ]` or `immarray [ ... ]`? How would efficiency compare between the `CollectionBuilder` approach extended to allow for CE-type behaviour, or hand-written builders? If the hand-written ones are better, could they be registered so they get implictly added when bare `[ ... ]` is interpreted as a collection?

Happypig375: This would need a comment from @T-Gro

RFCs/FS-1342-target-typed-literals.md
```diff
@@ -0,0 +1,186 @@
...
+2. `Span<T>` or `ReadOnlySpan<T>`;
+3. a type with a valid `CollectionBuilderAttribute`;
+4. a class or struct implementing `System.Collections.IEnumerable`, with an accessible applicable parameterless constructor and, for non-empty input, an accessible applicable instance or extension `Add`;
+5. `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`.
+
+Existing list construction remains available through ordinary checking. Multidimensional arrays and nullable lifting are outside the first version.
+
+Each target supplies one element type `T`. Expression elements are checked against `T`; a spread's iteration type must convert to `T`. Numeric and nested collection elements may themselves be contextual. The empty form requires a target that determines `T`.
+
+Source element and spread expressions are evaluated exactly once from left to right, before count queries or enumeration used to materialize the result.
```
T-Gro, line 99:
```suggestion
Source elements and `yield!` expressions are evaluated exactly once from left to right. Each `yield!` source is fully enumerated, including any required enumerator disposal, before the next element or `yield!` expression is evaluated.
```

Preserves existing F# evaluation order:

```fsharp
let shared = [|0|]
let mutate n = shared[0] <- n; shared

[yield! mutate 1; yield! mutate 2] // [1; 2]
```

Evaluating both sources before enumerating either produces `[2; 2]`. The suggestion avoids introducing that difference when contextual collection construction is selected.

RFCs/FS-1342-target-typed-literals.md
```diff
@@ -0,0 +1,193 @@
+# F# RFC FS-1342 - Contextual target typing for numeric and collection literals
+
+The design suggestion [Type inference for literals](https://github.com/fsharp/fslang-suggestions/issues/1421) is approved in principle. This RFC is the focused successor to [#800](https://github.com/fsharp/fslang-design/pull/800) and contains only levels 1–2 approved by its [design review](https://github.com/fsharp/fslang-design/pull/800#issuecomment-3852354596).
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/1421)
+- [x] Approved in principle
+- [ ] Implementation
+- [ ] [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)
+
+# Summary
```
T-Gro, line 10: I do like the framing from @charlesroddie of this being a compile-time improvement over op_Implicit conversions.

With the framing in mind (also, it would make sense if the table of conversions for numeric literals matches what FS-1093 provides - ideally with no important gaps), this could be a slimmer extension and provide more consistency to the language.

RFCs/FS-1342-target-typed-literals.md
```diff
@@ -0,0 +1,193 @@
...
+
+Ranges and collection-level `for`, `while`, `let`, `use`, `try`, `if`, or `match` retain their current list/sequence-expression meaning.
+
+An eligible expression may target:
+
+1. a one-dimensional array;
+2. `Span<T>` or `ReadOnlySpan<T>`;
+3. a type with a valid `CollectionBuilderAttribute`;
+4. a class or struct implementing `System.Collections.IEnumerable`, with an accessible applicable parameterless constructor and, for non-empty input, an accessible applicable instance or extension `Add`;
+5. `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`.
```
T-Gro, line 93: I know this is what C# supports, but we shall try to come up with a more elegant (less hardcoded) design that delegates special cases e.g. to library provided primitives.

Could maybe majority of cases be delegate to new FSharp.Core namespace and leave them to be library additions, while keeping the compiler addition smaller?

PR #839, opened 2026-08-27
RFC FS-1343: float64 abbreviation and d/D literal suffix

Happypig375:

## Summary

Adds two explicit names for the existing `System.Double` representation:

1. `float64` as an FSharp.Core type abbreviation, identical to `float` and `double`;
2. `d` and `D` as suffixes for ordinary decimal-form `System.Double` literals.

For example:

```fsharp
let x: float64 = 1d
let y = 1.25D

[<Measure>]
type s

let elapsed: float64<s> = 0.5d<s>
```

The RFC is independent of contextual or target-typed literals. It does not change:

- the default type of unsuffixed floating-point literals;
- the meaning of `float` or `double`;
- the runtime representation or emitted signatures;
- existing hexadecimal floating-point literal forms;
- units-of-measure behavior apart from making `float64<_>` an equivalent spelling.

The purpose is consistency with fixed-width integer names such as `int32` and `int64`, and explicit suffix symmetry with `float32` and `decimal`.

## Related work

- Split from section **e** of #800

This feature was approved separately in that review as an independent consistency improvement.

`RFCs/FS-1343-float64-and-d-suffix.md`

### DIFF
--- a/RFCs/FS-1343-float64-and-d-suffix.md
+++ b/RFCs/FS-1343-float64-and-d-suffix.md
@@ -1,8 +1,6 @@
 # F# RFC FS-1343 - `float64` abbreviation and `d`/`D` literal suffix
 
-This RFC is split from section **e** of [the original literal-inference RFC](https://github.com/fsharp/fslang-design/pull/800), as requested by [the design review](https://github.com/fsharp/fslang-design/pull/800#issuecomment-3852354596).
-
-The change is independent of target-typed literals: it adds an explicit spelling for the existing `System.Double` representation.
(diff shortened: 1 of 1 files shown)


RFCs/FS-1343-float64-and-d-suffix.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Type abbreviation
+
+FSharp.Core adds:
+
+```fsharp
+type float64 = System.Double
+
+[<MeasureAnnotatedAbbreviation>]
+type float64<[<Measure>] 'Measure> = float<'Measure>
```
T-Gro, line 39:
```suggestion
type float64<[<Measure>] 'Measure> = float<'Measure>
```

The attribute creates a distinct measured type; removing it makes `float64<s>` interchangeable with `float<s>`.

PR #840, opened 2026-08-27
RFC FS-1344: Deconstruct method support in tuple patterns

Happypig375:

## Summary

Allows an ordinary F# tuple pattern to consume a value whose independently known, non-tuple type exposes a compatible instance or extension `Deconstruct` method.

For example:

```fsharp
type Person(name: string, age: int) =
    member _.Deconstruct(nameResult: string outref, ageResult: int outref) =
        nameResult <- name
        ageResult <- age

let name, age = Person("Ada", 36)
```

The selected method follows the conventional CLI shape:

```csharp
void Deconstruct(out T1 item1, out T2 item2, ...)
```

Existing tuple inference remains the first-priority interpretation. When the pattern input type is not independently known, an ordinary tuple pattern continues to infer an ordinary F# tuple exactly as before.

`Deconstruct` lookup is used only after the input has already been established as a concrete non-tuple type. This avoids adding backwards inference or changing existing inferred function signatures.

The RFC specifies:

- eligible instance and extension methods;
- method shape and accessibility;
- arity and overload selection;
- ambiguity handling;
- evaluation and exception behavior;
- interaction with existing tuple, union-case, exception-case, and active patterns;
- diagnostics, navigation, and hover information;
- source and binary compatibility.

It adds no SRTP constraint, no generalized deconstruction constraint, and no field/property pattern facility.

## Related work

- Suggestion: https://github.com/fsharp/fslang-suggestions/issues/751
- Original omnibus RFC: #800

This PR is the focused successor to section **h** of #800.

`RFCs/FS-1344-deconstruct-patterns.md`

### DIFF
--- a/RFCs/FS-1344-deconstruct-patterns.md
+++ b/RFCs/FS-1344-deconstruct-patterns.md
@@ -1,0 +1,423 @@
+# F# RFC FS-1344 - `Deconstruct` method support in tuple patterns
+
+The design suggestion [Support C#-style Deconstruct method based pattern matching](https://github.com/fsharp/fslang-suggestions/issues/751) has been marked "approved in principle".
+
+This RFC is the focused successor to section **h** of [the original omnibus RFC](https://github.com/fsharp/fslang-design/pull/800), as requested by [the design review](https://github.com/fsharp/fslang-design/pull/800#issuecomment-3852354596).
+
+- [x] [Suggestion](https://github.com/fsharp/fslang-suggestions/issues/751)
+- [x] Approved in principle
+- [ ] Implementation
+- [ ] [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)
+
+# Summary
+
+Allow an ordinary F# tuple pattern to consume a value whose already-known, non-tuple type exposes a C#-compatible `Deconstruct` method.
+
+```fsharp
+type Person(name: string, age: int) =
+    member _.Deconstruct(nameResult: string outref, ageResult: int outref) =
+        nameResult <- name
+        ageResult <- age
+
+let name, age = Person("Ada", 36)
+```
+
+The selected method has the CLI shape:
+
+```csharp
+void Deconstruct(out T1 item1, out T2 item2, ...)
+```
+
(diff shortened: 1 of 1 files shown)


RFCs/FS-1344-deconstruct-patterns.md
```diff
@@ -0,0 +1,135 @@
...
+The proposal lets an existing .NET decomposition contract participate directly in nested patterns, without requiring a type-specific active-pattern adapter. It is especially useful for C# records, which synthesize `Deconstruct`.
+
+# Detailed design
+
+## Pattern selection
+
+For an ordinary tuple pattern with at least two components, apply the first matching case:
+
+1. **Input shape unresolved:** constrain it to an F# reference tuple, as today. Do not defer or revisit this choice.
+2. **Known reference tuple:** use existing tuple semantics and arity checking.
```
T-Gro, line 42:
```suggestion
2. **Known tuple, reference or struct:** use existing tuple semantics and arity checking.
```

An ordinary `(a, b)` pattern already accepts a known struct tuple, but the current list has no matching case.

RFCs/FS-1344-deconstruct-patterns.md
```diff
@@ -0,0 +1,135 @@
...
+
+A tested occurrence performs one call and creates one temporary per output, subject to semantics-preserving optimization. Candidate search uses ordinary member and extension lookup; pattern compilation otherwise scales as tuple-pattern compilation does today.
+
+## Culture-aware formatting/parsing
+
+Not applicable.
+
+# Unresolved questions
+
+Zero/one-output syntax and named positional subpatterns require separate proposals.
```
T-Gro, line 135:
```suggestion
Zero/one-output syntax and named positional subpatterns require separate proposals.

- For unannotated nonrecursive `let` bindings, which RHS type information is available before tuple fallback? How does this preserve the `Person` example and ordinary tuple patterns' influence on RHS overload resolution?
- Should ordinary `Deconstruct` lookup also require the compiled name `Deconstruct`? That additional condition excludes unrenamed optional F# extensions.
```

Current `let` checking starts with the pattern; the `Person` example needs RHS type information first. Unrenamed optional extensions emit `Type.Deconstruct`, so the name filter changes which extensions qualify.

PR #843, opened 2026-08-31
RFC FS-1347 - reraise in computation expressions

xperiandri:
RFC for the approved-in-principle suggestion [fslang-suggestions#660](https://github.com/fsharp/fslang-suggestions/issues/660): make `reraise ()` legal in the `with` handler of a `try ... with` written inside a computation expression, compiled as a rethrow through `ExceptionDispatchInfo` (same exception instance, stack trace preserved) since the handler is an ordinary function where IL `rethrow` is invalid.

Working implementation: dotnet/fsharp#20405 (preview feature `ReraiseInComputationExpressions`; async/task/custom builders and seq/list/array comprehensions; no FSharp.Core change).

Key decision points called out in the RFC's unresolved questions:
- closures escaping the handler (proposed: allowed — nearest-lexically-enclosing-handler rule);
- whether the `finally` position inside a handler should be excluded;
- guard-position `reraise` in sequence expressions runs inside a real CLR exception filter, where thrown exceptions are swallowed by the runtime — behavior documented, question raised whether it should be an error.

I picked FS-1347 as the next free number after the open RFC PRs; happy to renumber.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RFCs/FS-1347-reraise-in-computation-expressions.md
```diff
@@ -0,0 +1,144 @@
...
+## Motivation
+
+A computation expression's `try ... with` is translated to `builder.TryWith(builder.Delay(fun () -> body), function <clauses>)`. The handler is an ordinary function that the builder calls, not a .NET exception handler, so the IL `rethrow` instruction that `reraise ()` compiles to is not valid there.
+
+This forces the observability pattern above to be written as `raise (Exception("call failed", e))`, which allocates a new exception, changes the exception's type, buries the original one in `InnerException`, and resets the stack trace; or as `raise e`, which keeps the type but discards the original stack trace.
+
+.NET has provided `System.Runtime.ExceptionServices.ExceptionDispatchInfo` for exactly this situation since .NET Framework 4.5: it rethrows the original exception object with its stack trace preserved, from anywhere. The compiler already uses it for this exact reason — the unmatched clause of a computation expression handler compiles to `ExceptionDispatchInfo.Capture(e).Throw()` rather than `rethrow`. This RFC extends that treatment to `reraise ()` written by the user.
+
+## Detailed design
+
```
T-Gro, line 38: Please use the current template and keep the motivation at the user-visible level.

````suggestion
- [ ] [Discussion](https://github.com/fsharp/fslang-design/discussions/FILL-ME-IN)

# Summary

Permit `reraise ()` in computation expression handlers.

# Motivation

Computation expression handlers reject `reraise ()` with FS0413. Preserving exception identity and the original stack trace requires calling `ExceptionDispatchInfo.Capture(e).Throw()` explicitly.

```fsharp
async {
    try
        return! callService ()
    with e ->
        logger.LogError(e, "call failed")
        return reraise ()
}
```

# Detailed design

````

File: RFCs/FS-1347-reraise-in-computation-expressions.md [hunk 2]
```diff
@@ -0,0 +1,144 @@
...
+        raise Outer
+    with _ ->
+        try
+            raise Inner
+        with _ ->
+            reraise ()          // rethrows Inner, through IL 'rethrow'
+}
+```
+
+The rewrite requires the value the handler catches to have type `exn`. Builders whose `TryWith` takes a handler of any other type keep today's FS0413.
```
T-Gro, line 91: This states each semantic rule once and removes examples that duplicate the scope table.

```suggestion
## Semantics

In a computation expression handler, `reraise ()` refers to the exception passed to that handler throughout the lexical extent of its `with` clause. This applies whether or not the pattern names the exception.

Each invocation translates to `ExceptionDispatchInfo.Capture(e).Throw()`. It throws the same object, retains the existing stack trace, and adds an exception-dispatch boundary and rethrow frame. Direct-handler `reraise ()` keeps its existing IL `rethrow` semantics.

A closure can invoke `reraise ()` after its handler returns, repeatedly, or concurrently. Repeated calls can add exception-dispatch boundaries. Concurrent calls do not guarantee a deterministic trace shape.

In quotations, `reraise ()` produces a call to `ExceptionDispatchInfo.Capture(e).Throw()`.
```

T-Gro (commented):
RFC contract review.

RFCs/FS-1347-reraise-in-computation-expressions.md
```diff
@@ -0,0 +1,144 @@
...
+Positions that remain errors, unchanged:
+
+| Code | Diagnostic |
+|---|---|
+| `reraise ()` outside any handler | FS0413 |
+| `reraise ()` in the `try` body | FS0413 |
+| `let f = reraise` (first-class use) | FS0417 |
+| `reraise ()` in a builder whose handler does not take an `exn` | FS0413 |
+
+The `exn` requirement is checked at the moment the `reraise ()` call is type-checked, so a builder whose handler parameter type is generic or not yet solved at that point also keeps FS0413.
```
T-Gro, line 106: This makes the table the only scope reference and removes dependence on type-checking order.

```suggestion
## Applicability

The feature applies to builder-based computation expressions and to `seq`, list, and array expressions. The handler input type must be inferred as exactly `exn`.

| Position | Result |
|---|---|
| Handler body or guard, including after a bind | Allowed |
| Nested computation expression without a handler, local function, lambda, or `finally` | Inherits the enclosing computation expression handler |
| Protected body of a nested `try ... with` | Inherits the enclosing computation expression handler |
| `with` clause of a nested direct `try ... with` | Uses existing direct-handler rules and does not inherit the outer binding |
| `with` clause of a nested computation expression handler | Refers to that handler's exception |
| Outside any handler | FS0413 |
| Handler input type not inferred as exactly `exn` | FS0413 |
| First-class use such as `let f = reraise` | FS0417 |
```

RFCs/FS-1347-reraise-in-computation-expressions.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Known behavioral wrinkle: `when` guards in sequence expressions
+
+Sequence expressions compile their `try ... with` through the `RuntimeHelpers.EnumerateTryWith` runtime helper, which invokes the compiled guard from inside a real CLR exception filter block. The CLR discards exceptions thrown from a filter, so a guard-position `reraise ()` in a `seq { }` does not perform the `ExceptionDispatchInfo` rethrow: the original exception simply continues unwinding (which observationally resembles a rethrow, but appends no rethrow frame). In builder-based computation expressions, guard-position `reraise ()` behaves like any other position.
```
T-Gro, line 110: The current text incorrectly says the EDI rethrow does not execute. Only the feature-specific observable behavior is needed here.

```suggestion
In `seq`, list, and array expressions, a `when` guard that executes `reraise ()` runs once and skips subsequent clauses. It propagates the same exception with the guard frame and an exception-dispatch boundary.
```

RFCs/FS-1347-reraise-in-computation-expressions.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Compilation
+
+The type checker records the value holding the caught exception while checking the clause bodies and guards of a computation expression handler, and rewrites `reraise ()` against it. The handler of a real `try ... with` clears that record, which is what makes a nested real handler win.
+
+No FSharp.Core change is required — `ExceptionDispatchInfo` is a BCL type, and the compiler already resolves it for the unmatched-clause path, falling back to a plain `throw` on frameworks that lack it.
```
T-Gro, line 116: The RFC needs specification edits, not compiler bookkeeping.

```suggestion
# Changes to the F# spec

- **Expressions / Control Flow Expressions / Reraise Expressions**: permit `reraise ()` in the body and guards of a computation-expression or `seq`/list/array handler whose input type is inferred as exactly `exn`.
- **Expressions / Computation Expressions**: define exception identity, stack trace, lexical scope, quotation, and sequence-guard semantics.
```

RFCs/FS-1347-reraise-in-computation-expressions.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Relax the FS0413 check without changing the lowering.** Not possible — the IL `rethrow` instruction is invalid outside a handler block and would produce unverifiable code.
+
+## Compatibility
+
+- **Is this a breaking change?** No. Every program this accepts is one that previously failed to compile.
+- **What happens when previous versions of the F# compiler encounter this design addition as source code?** They report FS0413, as they do today.
+- **What happens when previous versions of the F# compiler encounter this design addition as compiled binaries?** Nothing; the feature leaves no trace in the assembly surface. The generated code calls only BCL methods.
+- **If this is a change or extension to FSharp.Core, what happens when previous versions of FSharp.Core are used by a compiler that supports this feature?** FSharp.Core is unchanged, so there is nothing to guard.
```
T-Gro, line 137: Please restore the required sections and record the non-obvious compatibility contract.

```suggestion
# Drawbacks

The computation-expression and direct-handler forms have different stack-trace shapes.

# Alternatives

- Keep FS0413 and require explicit exception threading.
- Lower to IL `rethrow`. This is invalid outside a CLR catch block.

# Prior art

C# preserves `throw;` across compiler-generated suspension in an `async` catch handler, but rejects it in nested lambdas and local functions.

# Compatibility

This is an additive, language-version-gated source feature. Earlier compilers report FS0413. Compiled code uses existing BCL calls and adds no FSharp.Core dependency.

The language-version check applies to producer source. A consumer using an earlier language version can inline a public definition compiled with this feature.

# Interop

Other .NET languages observe an ordinary throw of the same exception object. The feature adds no public API or metadata.

# Pragmatics

## Diagnostics

Earlier language versions report the standard language-version diagnostic.

## Tooling

No changes.

## Performance

One `ExceptionDispatchInfo` capture and throw per invocation.

## Scaling

Not applicable.

## Culture-aware formatting/parsing

Not applicable.
```

RFCs/FS-1347-reraise-in-computation-expressions.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Unresolved questions
+
+- Should `reraise ()` be allowed in closures that can escape the handler, or should the binding stop at user-written lambda boundaries? This RFC proposes allowing it, on the grounds that the rule "nearest lexically enclosing handler" is simpler to state and the escaping case is well-defined.
+- Under the same lexical rule, `reraise ()` in the `finally` of a `try ... finally` written inside a computation expression handler compiles and rethrows the caught exception, replacing any exception already in flight — a real handler rejects the identical shape with FS0413. Should the `finally` position be excluded?
+- Should guard-position `reraise ()` in sequence expressions be an error instead of the filter-swallowed behavior described above?
+- Should the FS0413 message point at the language version when the code would be accepted under `preview`?
```
T-Gro, line 144: These choices are specified above; keeping them open would contradict the proposed contract.

```suggestion
# Unresolved questions

None.
```

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro, line 106: The current implementation checks before inference completes, so a handler that later resolves to `exn` still gets FS0413. If this contract is accepted, defer the check and add a regression test.

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro, line 144: With `--langversion:11.0`, the implementation currently reports FS0413. If this contract is accepted, report the standard language-version diagnostic and add a regression test.

xperiandri:
Applied, including the two points listed as missing.

You were right about the sequence guard, and my wording was wrong. Measured on the built compiler: the guard runs once, the remaining clauses are skipped, and the exception that escapes carries the guard frame and the dispatch boundary, so the rethrow does execute.

Both implementation gaps are fixed in dotnet/fsharp#20405, with regression tests. The `exn` check now runs after inference, so a builder that leaves what its handler catches to inference qualifies once it resolves; an earlier language version reports FS3350 instead of FS0413.

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. The RFC now uses the current template, with a concise user-visible summary and motivation.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. The RFC now specifies that the guard runs once, skips subsequent clauses, and propagates the same exception with the guard frame and dispatch boundary.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. Applicability is consolidated into the scope table and the handler input must be inferred as exactly `exn`.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. The section now specifies the required F# specification changes without compiler bookkeeping.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. The choices are now part of the proposed contract and the RFC has no unresolved questions.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

T-Gro: Addressed in dotnet/fsharp#20405. Earlier language versions now report FS3350, with regression coverage for 11.0 and preview.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. The semantic rules are stated once, including closure lifetime, stack-trace behavior, direct-handler behavior, and quotations.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed in dotnet/fsharp#20405. The check is deferred until inference resolves the handler input type, with regression coverage for a type that later resolves to `exn`.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

RFCs/FS-1347-reraise-in-computation-expressions.md
T-Gro: Addressed. The required template sections and compatibility contract have been restored.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

